PR #6, opened 2024-10-30
Allow errorprone failures to be automatically suppressed

CRogers:
## Before this PR
Whenever a new errorprone check is rolled out at the ERROR level, it will cause failures on some of our many repos. The upgrade to the errorprones will be blocked until someone comes along to manually fix the error or suppress. Often this does not happen, so we end up with repos that have baseline upgrades blocked forever, which also blocks the gradle plugin from upgrading. Repos do not get the benefit of the new errorprone checks.

This is the big reason we do not make new errorprone checks, and if we do set them to WARNING instead of ERROR (which has no real value). 

## After this PR
==COMMIT_MSG==
Allow errorprone failures to be automatically suppressed
==COMMIT_MSG==

More details in the readme - please read. Also, the instructive comment in `SuppressibleErrorPronePlugin`.

### Overview

We achieve this with a two stage process:

1. Intercepting all the errors that errorprone produces, adding an `@RepeatableSuppressWarnings` annotation to the closest parent language element to erroring element that accepts `@SuppressWarnings`.
2. The second stage comes and coalesces all the `@RepeatableSuppressWarnings` together with any existing `@SuppressWarnings` to produce a final `@SuppressWarnings` (this process happens via a regular old errorprone check).

For example:

```java
class Example {
    @SuppressWarnings("ArrayToString")
    void example() {
        // Fails CollectionStreamForEach
        List.of(1).stream().forEach(...)
        // ...
    }
}
```

Would have the `@RepeatableSuppressWarnings` annotation added after stage 1:

```java
class Example {
    @RepeatableSuppressWarnings("CollectionStreamForEach")
    @SuppressWarnings("ArrayToString")
    void example() {
        List.of(1).stream().forEach(...)
        // ...
    }
}
```

Then stage 2 will coalesce the suppress warnings annotations into a single regular `@SuppressWarnings`. Note we prefix the automatically suppressed error with `for-rollout:` so it's easy to tell which suppressions happened because humans did it vs automation.

```java
class Example {
    @SuppressWarnings({"ArrayToString", "for-rollout:CollectionStreamForEach"})
    void example() {
        List.of(1).stream().forEach(...)
        // ...
    }
}
```

### How do we intercept all the errorprone errors?

We actually modify the core errorprone library to achieve this using a Gradle [Artifact Transform](https://docs.gradle.org/8.10.2/userguide/artifact_transforms.html). This allows us to minimally rewrite the bytecode in the jar that has [`VisitorState#reportMatch`](https://github.com/google/error-prone/blob/f0c3c1eb1b576ee9bc44f1f21c9379e7a02dd745/check_api/src/main/java/com/google/errorprone/VisitorState.java#L281) method and add a call to our own static method to modify the `description` as a first step and add our own fix for `@RepeatableSuppressWarnings`.

### How did we get that `for-rollout:` prefix to be accepted by errorprone

Again, we use the same Artifact Transform to also add a static method to the end of [this constructor for `BugCheckerInfo`](https://github.com/google/error-prone/blob/f0c3c1eb1b576ee9bc44f1f21c9379e7a02dd745/check_api/src/main/java/com/google/errorprone/BugCheckerInfo.java#L147) that reflectively changes the `allNames` field to include the check's canonical name with the `for-rollout:` prefix.

### What does this mean for baseline?

The bulk of the code for actually doing interesting errorprone related actions is now in this repo. [You can see the change for baseline here](https://github.com/palantir/gradle-baseline/blob/63b5336f9767a91b9dedcc4ccc5214533dcf0722/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java) ([diff](https://github.com/palantir/gradle-baseline/compare/callumr/use-suppressible-error-prone#diff-04c752ee8828edf8a30d4cd1eb21bdf7ee2451b2a0c17c352f57fd0988f24282)). `BaselineErrorProne` is essentially just configuration - is says which checks are on or off, and conditionally configures checks based on which libraries are used.

We already have more than one errorprone source (and I intend to add more). Baseline just becomes another errorprone source like any other.

### Future work

* Adding a way to remove unnecessary suppressions of error prone checks. This will be very useful when rolling an out an errorprone check that is incorrect and ends up spamming repos with suppressions. Just fixing the bug and upgrading should fix the bug.
* Automatically discovering errorprones to run from the manifest of jars. Will enable library authors to distribute fixes to upcoming breaking changes automatically.

## Possible downsides?
Using an artifact transform is kinda janky. If Google were to change these APIs or internal details, the code would break and we would have to fix (although I'm pretty sure we could - we could do whatever we want).

The ideal would be to upstream this into errorprone itself. Upstreaming may be possible, but this PR at least demonstrates this works without having to go through that process. See readme for the discussion about rolling our errorprones in monorepo vs polyrepo environments and how Google may not have our problem.


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,523 @@
...
+                }
+                
+                @SuppressWarnings("for-rollout:ArrayToString")
+                public static class SomeClass {
+                    static {
+                        System.out.println(new int[3].toString());
+                    }
+                }
+            }
+        '''.stripIndent(true)
```
CRogers, line 362: Shows the "suppress at every level" behaviour.

gradle-suppressible-error-prone/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+test {
+    systemProperty 'ignoreDeprecations', 'true'
+    systemProperty 'projectVersion', project.version
+}
+
+tasks.test.dependsOn tasks.findByPath(':suppressible-error-prone:publishToMavenLocal')
+tasks.test.dependsOn tasks.findByPath(':suppressible-error-prone-annotations:publishToMavenLocal')
```
felixdesouza: this seems a bit odd? can you not take a test dependency on `suppressible-error-prone`?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/ConditionalPatchCheck.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.suppressibleerrorprone;
+
+import java.util.Set;
+import org.gradle.api.specs.Spec;
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public final class ConditionalPatchCheck {
```
felixdesouza: not pressing, but you could actually make this a record if you so wished, similarly with `IfModuleIsUsed`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    private static void addAnnotationDependencyForSuppressingStage2(Project project, String version) {
+        if (isSuppressingStageTwo(project)) {
+            // Just before stage two of suppression starts compilation, we now have @RepeateableSuppressWarnings in
+            // the code, so we need to include the jar that has this type to each source set.
+            project.getExtensions().getByType(SourceSetContainer.class).configureEach(sourceSet -> {
+                project.getDependencies()
+                        .add(
+                                sourceSet.getCompileOnlyConfigurationName(),
```
felixdesouza, line 237: such awkward indentation, think it'll go away if you extracted `sourceSet.getCompile...` to a variable

suppressible-error-prone-annotations/build.gradle
```diff
@@ -0,0 +1,2 @@
+apply plugin: 'java-library'
```
esword, line 1: is this a separate project so that things outside of this repo can use the annotations if desired?  Or is it needed to add to the classpath of the target project itself when running the second pass?

readme.md
```diff
@@ -0,0 +1,104 @@
...
+
+dependencies {
+    // You will actually need a source of errorprone checks, this include the default Google ones
+    errorprone 'com.google.errorprone:error_prone_core:<version>'
+}
+
+suppressibleErrorProne {
+    patchChecks.add('SomeCheck')
+}
+```
```
esword, line 129: all the info you put in this PRs description would be useful in this readme under a "How does this work?" section as well.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/VisitorStateClassVisitor.java
```diff
@@ -0,0 +1,72 @@
...
+package com.palantir.gradle.suppressibleerrorprone.transform;
+
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+/**
+ * We want to intercept the error descriptions produced by errorprone checks that eventually go through `reportMatch`
+ * in `VisitorState` (even if the checks use the `buildDescription` approach). This is so we can modify the error
+ * descriptions to add our own fix. In the interests of minimising bytecode modifications, we just slot in a call
```
esword, line 26: minimizing... unless you brits spell it differently

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/ConditionalPatchCheck.java
```diff
@@ -0,0 +1,43 @@
+/*
```
esword, line 1: is this whole concept needed because something verifies that if you reference the name of a check in some configuration but that check can't be found on the classpath it throws an error?  I'm sketchy on the details of how checks are found/activated in our tooling so this may be a dumb question.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,261 @@
...
+                    .configure(ErrorProneOptions.class, errorProneOptions -> {
+                        configureErrorProneOptions(project, extension, javaCompile, errorProneOptions);
+                    });
+        });
+
+        addAnnotationDependencyForSuppressingStage2(project, version);
+
+        if (isAnyKindOfPatching(project)) {
+            project.afterEvaluate(_ignored -> {
+                // To allow refactoring near usages of deprecated methods, even when -Xlint:deprecation is specified,
```
esword, line 98: I understand what this is doing, but what does "near usages" mean?

esword (commented):
mostly just questions to verify I understand what's going on

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/VisitorStateClassVisitor.java
CRogers: 🇬🇧 🇬🇧 🇬🇧 🇬🇧 🇬🇧 🇬🇧 🇬🇧 🇬🇧 🇬🇧

CRogers: I accept using the American spellings in code for Java... comments are fair game though.

suppressible-error-prone-annotations/build.gradle
CRogers: The latter, see here:

https://github.com/palantir/suppressible-error-prone/blob/ee8fced1def59cbf7c37b414ae4c5fe4ea43e4f8/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L224-L234

readme.md
CRogers: Good point, I'll copy it over.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: I actually just [copied this from baseline](https://github.com/palantir/gradle-baseline/blob/4c71b30d2ba8a81cfb63cba8bd163207188c82a3/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java#L85-L100). I did actually change the comment though - it's not just refactoring of deprecated methods, it will fail if there are any usages of deprecated methods in the same compilation unit - see this test:

https://github.com/palantir/suppressible-error-prone/blob/ee8fced1def59cbf7c37b414ae4c5fe4ea43e4f8/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy#L406-L434

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/ConditionalPatchCheck.java
CRogers: For the safe `PreferSafeLogger` check for example, we only want to actually do automated refactoring from `Logger` to `SafeLogger` when the `com.palantir.safe-logger:logger` jar is on the classpath. We [already do this in baseline](https://github.com/palantir/gradle-baseline/blob/4c71b30d2ba8a81cfb63cba8bd163207188c82a3/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java#L260-L321), imo [it's much nicer with this construct](https://github.com/palantir/gradle-baseline/blob/63b5336f9767a91b9dedcc4ccc5214533dcf0722/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java#L63-L64).

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+                .getByName("jar")
+                .getAttributes()
+                .attribute(suppressible, false);
+
+        // It's the annotationProcessor configuration, not the errorprone that, is actually used by the compiler
+        // and so where we must put our transform. annotationProcessor extendsFrom errorprone.
+        project.getConfigurations().named("annotationProcessor").configure(errorProneConfiguration -> {
+            errorProneConfiguration
+                    .getDependencies()
+                    .add(project.getDependencies().create("com.google.errorprone:error_prone_check_api"));
```
felixdesouza, line 131: isn't this basically a no-op? shouldn't this be a dependency already?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -0,0 +1,158 @@
...
+        @Nested
+        protected abstract EnvironmentVariables getEnvironmentVariables();
+
+        public Params() {
+            // When running tests, you might want to debug why the artifact transform isn't working. Let me tell you,
+            // bashing your head against the wall because *for some reason* the artifact transform isn't running at
+            // all when it should is really frustrating, only to find out hours later that it's just getting cached
+            // as the inputs have not changed. So in tests only, we set the cache busting property to some random
+            // value so the transform always happens and you can actually debug the transform.
+            getCacheBust()
```
felixdesouza, line 148: that's odd, I would have thought changes to the transform would have cache bust the transform itself? would you need a general sort of cache busting for different releases?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private Optional<UnaryOperator<ClassVisitor>> classVisitorFor(String classJarPath) {
+        if (classJarPath.equals("com/google/errorprone/BugCheckerInfo.class")) {
+            return Optional.of(BugCheckerInfoVisitor::new);
+        }
+
+        if (classJarPath.equals("com/google/errorprone/VisitorState.class")
+                && getParameters().getSuppressionStage1().get()) {
+            return Optional.of(VisitorStateClassVisitor::new);
+        }
```
felixdesouza, line 91: I think this is a little strange. I don't think you need this as a parameter at all?

Instead just request different attributes for the dependency?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -0,0 +1,158 @@
...
+                try {
+                    InputStream entryInputStream = jarFile.getInputStream(jarEntry);
+                    Optional<byte[]> possibleNewClassBytes =
+                            classTransformer.transformClass(jarEntry.getName(), entryInputStream);
+
+                    if (possibleNewClassBytes.isPresent()) {
+                        byte[] newClassBytes = possibleNewClassBytes.get();
+
+                        jarEntry.setSize(newClassBytes.length);
+                        jarEntry.setCompressedSize(-1);
```
felixdesouza, line 114: what's the deal here?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -0,0 +1,158 @@
...
+        // https://github.com/gradle/gradle/issues/11831
+        if (inputName.startsWith("error_prone_check_api")) {
+            suppressCheckApi(outputs.file(inputName.replace(".jar", "") + "_suppressible_error_prone_modified.jar"));
+            return;
+        }
+
+        // There's no way to just run the transform on a single jar/component:
+        // https://github.com/gradle/gradle/issues/8386
+        // Instead we run the transform on every jar but just don't modify the jars we don't care about.
+        outputs.file(getInputArtifact());
```
felixdesouza, line 62: does that do a copy of all the jars? can you instead just return instead?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+            errorProneConfiguration.getAttributes().attribute(suppressible, true);
```
felixdesouza, line 132: wouldn't it be sufficient to put `isSuppressionStageOne` here?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
felixdesouza: I wonder whether you could put an attribute on the dependency itself

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+            });
+            return;
+        }
+
+        if (isApplyingSuggestedPatches(project)) {
+            errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
+                @Override
+                public Iterable<String> asArguments() {
+                    String possibleSpecificPatchChecks = (String) project.property(ERROR_PRONE_APPLY);
+                    if (!(possibleSpecificPatchChecks == null || possibleSpecificPatchChecks.isBlank())) {
```
felixdesouza: nit: can you push through the negation?

```java
possibleSpecificPatchChecks != null && !possibleSpecificPatchChecks.isBlank()
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+                        List<String> specificPatchChecks = Arrays.stream(possibleSpecificPatchChecks.split(","))
+                                .map(String::trim)
+                                .filter(Predicate.not(String::isEmpty))
+                                .collect(Collectors.toList());
```
felixdesouza: you can just do `toList()` now in java 17

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+                    // The behaviour of "-XepPatchChecks:" is to patch *all* checks that are enabled, so we can't
+                    // just leave it as that.
+                    if (patchChecks.isEmpty()) {
+                        return List.of();
+                    }
+
+                    return List.of("-XepPatchLocation:IN_PLACE", "-XepPatchChecks:" + String.join(",", patchChecks));
+                }
+            });
+            return;
```
felixdesouza: this seems redundant?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    private static void addAnnotationDependencyForSuppressingStage2(Project project, String version) {
+        if (isSuppressingStageTwo(project)) {
```
felixdesouza, line 231: is there any harm in just always adding it?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -0,0 +1,93 @@
...
+        switch (kind) {
+            case CLASS:
+            case METHOD:
+            case VARIABLE:
+                // VARIABLE includes fields
+                return true;
+            default:
+                return false;
+        }
+    }
```
felixdesouza: you can do switch expressions now!:

```java
return switch (kind) {
  case CLASS, METHOD, VARIABLE -> true;
  default -> false;
}
```

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RefactoringValidator.java
```diff
@@ -0,0 +1,115 @@
...
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.CompilationTestHelper;
+import com.google.errorprone.annotations.CheckReturnValue;
+import com.google.errorprone.bugpatterns.BugChecker;
+
+/**
+ * {@link RefactoringValidator} delegates to a {@link BugCheckerRefactoringTestHelper}, but also validates the output
+ * passes validation.
+ */
+final class RefactoringValidator {
```
felixdesouza, line 31: nice!

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesceTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .doTest();
+    }
+
+    @Test
+    void single_suppress_warnings_multiple_repeatable_warnings() {
+        fix().addInput(
+                        "Test.java",
+                        // language=Java
+                        """
+                        public class Test {
```
felixdesouza, line 81: text blocks make this so much nicer 😍

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesce.java
```diff
@@ -0,0 +1,160 @@
...
+        if (annotationType instanceof MemberSelectTree) {
+            return ((MemberSelectTree) annotationType).getIdentifier();
+        }
+
+        throw new UnsupportedOperationException(
+                "Unsupported annotation type: " + annotationType.getClass().getCanonicalName());
+    }
+
+    private static Stream<String> annotationStringValues(AnnotationTree annotation) {
+        return annotation.getArguments().stream().flatMap(arg -> {
```
felixdesouza, line 133: nit: `filter`?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesce.java
```diff
@@ -0,0 +1,160 @@
...
+        return buildDescription(tree)
+                .setMessage("Coalescing @SuppressWarnings annotations")
+                .addFix(fixBuilder.build())
+                .build();
+    }
+
+    private static Name annotationName(Tree annotationType) {
+        if (annotationType instanceof IdentifierTree) {
+            return ((IdentifierTree) annotationType).getName();
+        }
```
felixdesouza, line 122: you should be able to add a variable and refer to that without having to cast anymore!

```java
if (annotaitonType instanceof IdentifierTree identifierTree) {
  return identifierTree.getName();
}
```

etc.

You could even do it via a switch if you were so inclined:

```java
switch (annotationType) {
  case IdentifierTree identifierTree -> identifierTree.getName();
  case MemberSelectTree memberSelectTree -> memberSelectTree.getIdentifier();
  default -> throw new UnsupportedOperationException(...);
```

felixdesouza: actually, switch pattern matching is only java 21, ignore me

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesce.java
felixdesouza: or alternatively, I think you can do:

```java
if (!arg instanceof AssignmentTree assignment) {
  return Stream.empty();
}

if (assignment.getExpression() instanceof LiteralTree literal) {
  return Stream.of(literal.getValue());
}

if (assignment.getExpression() instanceof NewArrayTree newArray) {
  return ...
}

throw new UnsupportedOperationException(...);
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesce.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.lang.model.element.Name;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "blah")
```
felixdesouza: nit:

gradle-suppressible-error-prone/src/test/java/com/palantir/gradle/suppressibleerrorprone/ExcludedPathsRegexTest.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+final class ExcludedPathsRegexTest {
+    private static final Pattern PATTERN = Pattern.compile(SuppressibleErrorPronePlugin.excludedPathsRegex());
+
+    @ParameterizedTest(name = "{0}: {1}")
```
felixdesouza, line 30: take me back to spock 😭

felixdesouza: I mean, in fairness, surely it's easier to just inline this completely?

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+package com.palantir.gradle.suppressibleerrorprone
+
+import nebula.test.IntegrationSpec
+import nebula.test.functional.ExecutionResult
+import org.apache.commons.io.FileUtils
+
+class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
+    // We need to put the source sets in a different directory that does not contain the any words that would hit
+    // the errorprone excludedPathRegex, ie build in build/nebulatest
+    static File nebulatestSourceSets = new File('nebulatestSourceSets/' + SuppressibleErrorPronePluginIntegrationTest.class.simpleName)
```
felixdesouza, line 26: oof

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+            package app;
+            public final class App {
+                public static void main(String[] args) {
+                    System.out.println(new int[3].toString());
+                }
+            }
+        '''.stripIndent(true)
+
+        when:
+        runTasksSuccessfully('compileJava', '-PerrorProneApply')
```
felixdesouza, line 202: why was this successful? shouldn't it fail?

felixdesouza: or is apply just fix any thing that's fixable and leave the rest alone?

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+            }
+        '''.stripIndent(true)
+
+        when:
+        // Doesn't actually do any patching as the set is empty. It just does a normal compile that fails.
+        def stderr = runTasksWithFailure('compileJava', '-PerrorProneApply').standardError
+
+        then:
+        stderr.contains('[ArrayToString]')
+        appJava.text.contains('new int[3].toString()')
```
felixdesouza, line 232: I feel like this is confusing behaviour from my above comment, it'll likely not matter as this isn't exactly for typical consumers

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+                }
+            }
+        '''.stripIndent(true)
+
+        then:
+        runTasksSuccessfully('compileJava', '-PerrorProneDisable')
+        runTasksSuccessfully('compileJava', '-Pcom.palantir.baseline-error-prone.disable')
+    }
+
+    def 'should be able to refactor near usages of deprecated methods'() {
```
felixdesouza, line 417: what actually goes wrong/what are you fixing here, I'm not aware

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+
+    def 'IfModuleIsUsed works properly'() {
+        // language=Gradle
+        buildFile << '''
+            import com.palantir.gradle.suppressibleerrorprone.ConditionalPatchCheck
+            import com.palantir.gradle.suppressibleerrorprone.IfModuleIsUsed
+
+            suppressibleErrorProne {
+                conditionalPatchChecks.add(new ConditionalPatchCheck(new IfModuleIsUsed('com.fasterxml.jackson.core', 'jackson-core'), 'ArrayToString'))
+                conditionalPatchChecks.add(new ConditionalPatchCheck(new IfModuleIsUsed('donesnt', 'exist'), 'ArrayEquals'))
```
felixdesouza:
```suggestion
                conditionalPatchChecks.add(new ConditionalPatchCheck(new IfModuleIsUsed('doesnt', 'exist'), 'ArrayEquals'))
```

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,531 @@
...
+        appJava = file('app/App.java', mainSourceSet)
+
+        // language=Gradle
+        buildFile << '''
+            apply plugin: 'com.palantir.suppressible-error-prone'
+            apply plugin: 'java'
+            
+            repositories {
+                mavenCentral()
+                mavenLocal()
```
felixdesouza, line 50: `mavenLocal()` ?

felixdesouza (commented):
very exciting! lots of questions

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
CRogers: It's more if you've already written the gradle code, then you:

1. Run a test: it runs the transform.
2. Test fails, you rerun the test in debug mode.
3. Your breakpoint in the transform code is never hit as the transform is checked.
4. You think the transform isn't running at all.

If you change the actual code enough, it will actually rerun. But eventually it'll silently start skipping again.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: I tried. I even asked in the Gradle slack. It's not possible, there's [a Gradle issue for it](https://github.com/gradle/gradle/issues/8386):

https://github.com/palantir/suppressible-error-prone/blob/5797eafcb788f68871bbe7292e165ce4eed71524/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java#L59-L62

CRogers: > isn't this basically a no-op? shouldn't this be a dependency already?

The errorprone plugin does not add any dependencies by default to the errorprone configuration. Makes sense, maybe you don't want the default google errorprones.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
CRogers: I want to ensure we have an unmodified errorprone codepath for the `reportMatch` path if we're not suppressing. I could use different attributes and different transforms, but I think this would be worse. Since there's so much code shared between the the transforms, I'd probably make a helper classes, or subclasses that are also parameterised. Why not just use the built in artifact transforms parameterisation instead?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
CRogers: This is just what you do when adding stuff to zip via this api. You can't give it the compressed size ahead of time, it recalculates it if it's not valid

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
CRogers: Just returning means there's no jar in the transform output.

This doesn't do a copy. It just tells the artifact transform it's not been transformed. From the compiler `annotationProcessor` classpath, compare `error_prone_check_api` which is transformed:

```
/Users/callumr/.gradle/caches/8.8/transforms/ebe834d660cd113be831b2117daf4e2e/transformed/error_prone_check_api-2.31.0_suppressible_error_prone_modified.jar
```

to `error_prone_annotation` which is not:

```
/Users/callumr/.gradle/caches/modules-2/files-2.1/com.google.errorprone/error_prone_annotation/2.31.0/6eaafd66f2517268792ea54dddd17176f4008bd/error_prone_annotation-2.31.0.jar
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: We still need to run the transform for the `for-rollout:` naming prefix in `BugCheckerInfo` even when running normal compilation.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: Done, and extracted a var.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: nice

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: People would be able to use it in code... I don't want it exposed if we don't need to.

CRogers: Added a comment

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: More PJF issues lol:

Screenshot 2024-11-04 at 14 00 01

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: Apply just applies fixes, it doesn't fail if there are errors left over.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsCoalesce.java
CRogers: `if (!arg instanceof AssignmentTree assignment` is illegal apparently, and doesn't make a reference outside of the if. I was thinking that would be some crazy scoping rules if it did.

CRogers: actually, it does work, changed.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: Added a comment

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: Unfortunately not fixed by that.

gradle-suppressible-error-prone/build.gradle
CRogers: These get added to the compiler classpath and compileOnly of sourcesets, both inside nebula-test gradle runs, so not use in the actual test classpath.

CRogers: Added comment.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: We do need this, added comment

felixdesouza: so if I've understood correctly, you want to add a dependency without any shenanigans on the `suppressible-error-prones` piece within a gradle test, and only way to do that nicely is through `mavenLocal` ?

It'll be on the test class path under libs, but not something accessible to the build is what I'm guessing

CRogers: Yep

felixdesouza:
👍  pending comment

felixdesouza (approved):

PR #8, opened 2024-11-05
Support non-main source sets

CRogers:
## Before this PR
In #6, we allowed errorprones to be automatically suppressed. However, this only worked for the `main` source set. Any other sourceset (like `test`) did not get this feature.

## After this PR
==COMMIT_MSG==
* Support non-main source sets (eg `test`)
* Provide a `compileAllErrorProne` task that depends on all the compile tasks with that have errorprone applied and enabled.
==COMMIT_MSG==

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorProneExtension.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorProneExtension.java
@@ -22,7 +22,6 @@ import java.util.stream.Stream;
 import net.ltgt.gradle.errorprone.ErrorProneOptions;
 import org.gradle.api.Action;
 import org.gradle.api.Project;
-import org.gradle.api.plugins.ExtensionAware;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.SetProperty;
 import org.gradle.api.tasks.compile.JavaCompile;
@@ -49,9 +48,9 @@ public abstract class SuppressibleErrorProneExtension {
     }
 
     public final void configureEachErrorProneOptions(Action<ErrorProneOptions> action) {
-        project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> ((ExtensionAware)
-                        javaCompile.getOptions())
-                .getExtensions()
-                .configure(ErrorProneOptions.class, action));
+        project.getTasks()
+                .withType(JavaCompile.class)
+                .configureEach(
+                        javaCompile -> SuppressibleErrorPronePlugin.configureErrorProneOptions(javaCompile, action));
     }
 }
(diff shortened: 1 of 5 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -110,6 +110,12 @@ private void applyToJavaProject(Project project) {
                 });
             });
         }
+
+        project.getTasks().register("compileAllErrorProne", Task.class, compileAll -> {
+            compileAll.dependsOn(project.getTasks().withType(JavaCompile.class).matching(javaCompile -> {
+                return errorProneOptionsFor(javaCompile).getEnabled().get();
```
felixdesouza: this is a pitfall, if enablement changes at some later point, you'll never get these in your filtered `TaskCollection`.

I *think* you might want to wrap it in a provider perhaps? such that it's evaluated when `dependsOn` needs it, vs in configuration?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorProneExtension.java
```diff
@@ -49,9 +48,9 @@ public final Set<String> patchChecksForCompilation(JavaCompile javaCompile) {
     }
 
     public final void configureEachErrorProneOptions(Action<ErrorProneOptions> action) {
-        project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> ((ExtensionAware)
-                        javaCompile.getOptions())
-                .getExtensions()
-                .configure(ErrorProneOptions.class, action));
+        project.getTasks()
```
felixdesouza, line 51: much nicer!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: I thought `matching` was lazy? It really hope it is at least. I guess if it's not my test could be working because `compileAllErrorProne` has not be realised by the time we disable error prone...

felixdesouza: hmm, you *may* be okay, perhaps I'm overfitting due to what I had to deal with in an internal project. It is lazy, it just depends when you that collection is iterated upon, so I imagine it *should* be fine as long as you don't try to do anything with it during configuration?

felixdesouza:
👍

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: I'm just going to add a provider just in case

felixdesouza:
👍

PR #9, opened 2024-11-05
Set tags too to try to fix Gradle plugin portal publish

CRogers:
## Before this PR
Publish is still failing as [we didn't set `tags`](https://app.circleci.com/pipelines/github/palantir/suppressible-error-prone/61/workflows/d1536002-97e2-4bc0-9126-4a133710febb/jobs/111?invite=true#step-104-18424_60).

## After this PR
==COMMIT_MSG==
Set tags too to try to fix Gradle plugin portal publish
==COMMIT_MSG==


felixdesouza:
👍

PR #14, opened 2024-11-06
Don't fail to suppress on non-class type definitions (interfaces, enums, records etc)

CRogers:
## Before this PR
Suppressing fails when there is an errorprone error raised on an interface, enum, record etc (anything that isn't a class):

```
public interface InterfaceType {
       ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.31.0
     BugPattern: SafeLoggingPropagation
     Stack Trace:
     java.lang.NullPointerException: Cannot invoke "com.sun.source.util.TreePath.getLeaf()" because "path" is null
        at com.palantir.suppressibleerrorprone.VisitorStateModifications.lambda$interceptDescription$0(VisitorStateModifications.java:44)
```

This is because we never find a top level type we can suppress, as we didn't think of anything but class as suppressible.

## After this PR
==COMMIT_MSG==
Don't fail to suppress on non-class type definitions (interfaces, enums, records etc)
==COMMIT_MSG==

I've also significantly improved the error message when this does go wrong:

```
public interface InterfaceType {
       ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.31.0
     BugPattern: SafeLoggingPropagation
     Stack Trace:
     java.lang.RuntimeException: Can't find any source element on the TreePath to the error to place a @SuppressWarnings on. This is a bug an issues with suppressible-error-prone. The path to the error is:
  
  ===========================
  INTERFACE
  ===========================
  
  public interface InterfaceType {
      // ...
  }
  
  ===========================
  COMPILATION_UNIT
  ===========================
  package app;
  
  // ...
  
  public interface InterfaceType {
      // ...
  }
        at com.palantir.suppressibleerrorprone.VisitorStateModifications.lambda$interceptDescription$3(VisitorStateModifications.java:57)

```

### DIFF
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
@@ -52,12 +52,18 @@ public final class VisitorStateModifications {
                 .dropWhile(path -> !suppressibleKind(path.getLeaf().getKind()))
                 .findFirst()
                 .orElseThrow(() -> {
-                    return new RuntimeException("Can't find any source element on the TreePath to the error to "
-                            + "place a @SuppressWarnings on. This is a bug with suppressible-error-prone. "
-                            + "The path to the error is:\n\n"
-                            + StreamSupport.stream(pathToActualError.spliterator(), false)
-                                    .map(tree -> tree.getKind().name() + "\n===========================\n" + tree)
-                                    .collect(Collectors.joining("\n\n")));
+                    return new RuntimeException(
+                            """
+                            Can't find any source element on the TreePath to the error to \
+                            place a @SuppressWarnings on. This is a bug with suppressible-error-prone.
+                            The path to the error is:
+
+
+                            """
+                                    + StreamSupport.stream(pathToActualError.spliterator(), false)
+                                            .map(tree ->
+                                                    tree.getKind().name() + "\n===========================\n" + tree)
+                                            .collect(Collectors.joining("\n\n")));
                 })
                 .getLeaf();
 
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
@@ -53,7 +53,7 @@ public final class VisitorStateModifications {
                 .findFirst()
                 .orElseThrow(() -> {
                     return new RuntimeException("Can't find any source element on the TreePath to the error to "
-                            + "place a @SuppressWarnings on. This is a bug an with suppressible-error-prone. "
+                            + "place a @SuppressWarnings on. This is a bug with suppressible-error-prone. "
                             + "The path to the error is:\n\n"
                             + StreamSupport.stream(pathToActualError.spliterator(), false)
                                     .map(tree -> tree.getKind().name() + "\n===========================\n" + tree)
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
@@ -19,6 +19,7 @@ package com.palantir.suppressibleerrorprone;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.util.TreePath;
 import com.sun.tools.javac.util.JCDiagnostic.DiagnosticPosition;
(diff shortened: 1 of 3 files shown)


suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -44,10 +47,18 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
                 .dropWhile(path -> !suppressibleKind(path.getLeaf().getKind()))
                 .findFirst()
-                .orElseThrow(() -> new RuntimeException("Can't find anything we can suppress"))
+                .orElseThrow(() -> {
+                    return new RuntimeException("Can't find any source element on the TreePath to the error to "
+                            + "place a @SuppressWarnings on. This is a bug an with suppressible-error-prone. "
+                            + "The path to the error is:\n\n"
+                            + StreamSupport.stream(pathToActualError.spliterator(), false)
+                                    .map(tree -> tree.getKind().name() + "\n===========================\n" + tree)
+                                    .collect(Collectors.joining("\n\n")));
```
felixdesouza: text block? 🥺

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -44,10 +47,18 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
 
-        Tree firstSuppressibleParent = Stream.iterate(pathToActualError, TreePath::getParentPath)
+        Tree firstSuppressibleParent = Stream.iterate(
+                        pathToActualError, treePath -> treePath.getParentPath() != null, TreePath::getParentPath)
                 .dropWhile(path -> !suppressibleKind(path.getLeaf().getKind()))
                 .findFirst()
-                .orElseThrow(() -> new RuntimeException("Can't find anything we can suppress"))
+                .orElseThrow(() -> {
+                    return new RuntimeException("Can't find any source element on the TreePath to the error to "
+                            + "place a @SuppressWarnings on. This is a bug an with suppressible-error-prone. "
```
felixdesouza:
```suggestion
                            + "place a @SuppressWarnings on. This is a bug with suppressible-error-prone. "
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: I thought maybe but I don't think it's worth the bad formatting that palantir-java-format does without having actual newlines in the middle of the text.

felixdesouza: i.e.

```
new RuntimeException("""
  Can't find any source element on the TreePath to the error to place a @SuppressWarnings on. This is a bug with suppressible-error-prone.

  The path to the error is:

  %s

  ===================

  %s
  """.formatted(StreamSupport.stream(...), tree);

potentially translated wrong but I think it works

Comment (CRogers) [reply suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java]: We should probably fix PJF for this case:

```java
"""
Can't find any source element on the TreePath to the error to \
place a @SuppressWarnings on. This is a bug an with suppressible-error-prone.
The path to the error is:


"""
        + StreamSupport.stream(pathToActualError.spliterator(), false)
                .map(tree ->
                        tree.getKind().name() + "\n===========================\n" + tree)
                .collect(Collectors.joining("\n\n")));
```

should be

```java
"""
Can't find any source element on the TreePath to the error to \
place a @SuppressWarnings on. This is a bug an with suppressible-error-prone.
The path to the error is:


""" + StreamSupport.stream(pathToActualError.spliterator(), false)
        .map(tree ->
                tree.getKind().name() + "\n===========================\n" + tree)
        .collect(Collectors.joining("\n\n")));
```

imo

CRogers: It also newlines + double indents `.formatted`.

felixdesouza: true, but I would just do `"""...""".formatted(StreamSupport.stream())` and/or extract that to a variable. It is quite a gnarly expression to tack on

felixdesouza: > It also newlines + double indents `.formatted`.

oof

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: fixed

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: Eh, I've done the text block version so if we fix PJF it should get autoformatted to something nice.

felixdesouza (approved):

PR #22, opened 2024-11-14
Publish libraries that are run during compilation at target level 11

CRogers:
## Before this PR
Unfortunately, there's a reasonable number of projects that still use 11 for compilation. The rollout of gradle-baseline 6.2.0 was blocked by having this code that runs inside compilation at target level 17.

## After this PR
==COMMIT_MSG==
Libraries from this repo that are used in compilation (`suppressible-error-prone`, `suppressible-error-prone-annotations`) are now targeted at JDK 11 rather than 17.
==COMMIT_MSG==

## Possible downsides?
* Custom overrides that will need to be removed at some point.
* Can't use nice source features like text blocks.


felixdesouza:
👍

PR #27, opened 2024-11-21
Tests for suggestion and warning levels

CRogers:
## Before this PR
Had a minor panic about the changes made in OSS. Turns out they were all WARNINGs, so probably fine. But not tests for these cases

## After this PR
Add tests for these cases.


felixdesouza:
👍

PR #29, opened 2024-11-22
Handle errors on import statements

CRogers:
## Before this PR
Currently, if there is an error on an import, suppression will fail (because there is no way to place a `@SuppressWarnings`). People shouldn't make `ImportTreeMatcher`s that are at error level or do not have an always succeeding autofixer as they can't be suppressed.

## After this PR
==COMMIT_MSG==
Do not fail to suppress with an exception if there is an error on an import statement.
==COMMIT_MSG==

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -670,6 +670,50 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         '''.stripIndent(true)
     }
 
+    def 'makes no changes when there is an error on an import'() {
+        when: 'theres an illegal import'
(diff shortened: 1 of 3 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -670,6 +670,50 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
...
+        then: 'nothing has changed as we cant put SuppressWarnings on an import'
+        def stderr2 = runTasksWithFailure('compileAllErrorProne').standardError
+        stderr2.contains('[NonCanonicalStaticImport]')
+
+        // language=Java
+        appJavaTextEquals '''
+            package app;
+            import static app.B.Inner;
+            public final class App {}
+        '''.stripIndent(true)
```
felixdesouza, line 714: would it be clearer to save the output of lines 691-695 to a variable, and say that it is still equal?

felixdesouza:
👍

PR #39, opened 2025-01-16
Suppress in one pass for better performance

CRogers:
## Before this PR
Currently, suppressible-error-prone requires you to do two compilations in order to suppress errors:

```
./gradlew compileAllErrorProne -PerrorProneSuppressStage1
./gradlew compileAllErrorProne -PerrorProneSuppressStage2
```

Not only is this more complicated to use (and maintain) than a single pass, it increases the time it takes to do suppress. When combined with applying fixes first, it actually requires three compilations:

```
./gradlew compileAllErrorProne -PerrorProneApply
./gradlew compileAllErrorProne -PerrorProneSuppressStage1
./gradlew compileAllErrorProne -PerrorProneSuppressStage2
```

Excavator checks have a time limit, and some internal projects are getting close to or exceeding the limit by having to repeat compilation so many times.

Additionally, the concept of `@RepeatableSuppressWarnings` needs to be understood by maintainers, and these end up appearing if stage 1 fails.

## After this PR
==COMMIT_MSG==
Suppression is now just a one step process rather than two commands:

```
./gradlew compileAllErrorProne -PerrorProneSuppress
```
==COMMIT_MSG==

This makes usage much simpler and it's now faster to suppress errorprones, with less understanding overhead.

The way we do this also opens up the possibility of doing suggested refactoring in the same compile in a future PR, saving even more time:

```
./gradlew compileAllErrorProne -PerrorProneApply -PerrorProneSuppress
```

If we build the "remove unnecessary errorprone suppressions" feature too, previously would have required another two compilations (bringing it up to five for apply, suppress, remove unnecessary), but could just end up as one compilation still:

```
./gradlew compileAllErrorProne -PerrorProneApply -PerrorProneSuppress -PerrorProneRemoveUnnecessarySuppresions
```

## Possible downsides?
* We have to store some extra state in a weak hash map to do this in one pass, which is kinda odd
* We have to ensure we're lazy enough at the Fix/Replacement level to work correctly with errorprone.


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/AbstractSuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,129 @@
...
+ */
+
+package com.palantir.gradle.suppressibleerrorprone
+
+
+import nebula.test.IntegrationSpec
+import nebula.test.functional.ExecutionResult
+import org.apache.commons.io.FileUtils
+
+class AbstractSuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
```
CRogers: I separated out the current test file into two test with an abstract class for the common logic:
1. Tests that interact with the gradle errorprone options
2. Checked-in style tests that test how suppressing works

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/CodeChangesTest.groovy
```diff
@@ -0,0 +1,217 @@
...
+import com.google.common.base.Splitter
+import one.util.streamex.StreamEx
+import spock.lang.Unroll
+
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.stream.Collectors
+import java.util.stream.Stream
+
+class CodeChangesTest extends AbstractSuppressibleErrorPronePluginIntegrationTest {
```
CRogers: Best to look at [`AddNewSuppressionsAtEveryLevel.java`](https://github.com/palantir/suppressible-error-prone/pull/39/files#diff-30997f42a0f6268796064a71edc68fd4810b56ce453258fe040218d8da3d163f) and on to get an idea of what the test for this look like

gradle-suppressible-error-prone/src/test/resources/suppression-test/AddNewSuppressionsAtEveryLevel.java
```diff
@@ -0,0 +1,53 @@
+// Args: -PerrorProneSuppress
```
CRogers: This will allow us to later test running `-PerrorProneApply -PerrorProneSuppress` together and check it does the right thing.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/AbstractSuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -0,0 +1,129 @@
...
+                        @Override
+                        public Iterable<String> asArguments() {
+                            return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5005")
+                        }
+                    })
+                }
+            }
+        '''.stripIndent(true)
+
+        buildFile << """
```
esword: why are these lines here vs in the source-sets block above?  just because having to do variable resolution?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -63,60 +75,76 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
-        return Optional.ofNullable(visitorState.getSourceCode())
-                .map(sourceCode -> whitespaceIndentBefore(
-                        sourceCode, ((DiagnosticPosition) firstSuppressibleParent).getStartPosition()));
+    private static boolean suppressibleTree(Tree tree) {
+        return isModifiersTree(tree).isPresent();
     }
 
-    static CharSequence whitespaceIndentBefore(CharSequence sourceCode, int sourceElementPosition) {
-        int pos = sourceElementPosition - 1;
+    private static Optional<ModifiersTree> isModifiersTree(Tree tree) {
```
esword: method name a little strange since it doesn't return a boolean.  maybe getModifiersTree or just getModifiers

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -63,60 +75,76 @@ public static Description interceptDescription(VisitorState visitorState, Descri
 
         Tree firstSuppressibleParent = firstSuppressible.get().getLeaf();
 
-        // Guess the indent if we can't find it for some reason. Formatter will fix.
-        CharSequence indent =
-                indentForTree(visitorState, firstSuppressibleParent).orElse("    ");
+        ModifiersTree modifiersTree = isModifiersTree(firstSuppressibleParent).get();
```
esword: since modifiersTree is only used to find the suppressWarnings annotation, just wrap this and the next block in a method with a useful name - e.g. getSuppressWarningsAnnotation.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
```diff
@@ -0,0 +1,208 @@
...
+    public boolean isEmpty() {
+        return false;
+    }
+
+    private static final class SuppressingReplacement extends Replacement {
+        // We really do need to be this lazy for generating the Replacements, as error-prone immediately converts
+        // the Fix to a Replacement when a Description is given to it, and we need to defer the computation of the
+        // Replacement until a number of Descriptions have been produced, to handle multiple errors being suppressed
+        // at the same level.
+        private final Supplier<Replacement> replacement;
```
esword, line 174: does this need to be a caching supplier?  i.e. does calling the replacement.get() multiple times cause issues, either with perf or by calling fix() multiple times which then calls fixWithExistingSuppressWarnings multiple times which then... ?  My brain is starting to hurt so I'm not sure there is any problem.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
```diff
@@ -0,0 +1,208 @@
...
+                annotationStringValues(suppressWarningsAnnotation).collect(Collectors.toSet());
+
+        Set<String> toAdd = errors.stream()
+                .filter(Predicate.not(error -> existingValues.contains(error)
+                        || existingValues.contains(CommonConstants.AUTOMATICALLY_ADDED_PREFIX + error)))
+                .collect(Collectors.toSet());
+
+        List<String> warningsToSuppress = Stream.concat(
+                        existingValues.stream(),
+                        toAdd.stream().sorted().map(warning -> CommonConstants.AUTOMATICALLY_ADDED_PREFIX + warning))
```
esword: nit: this sorts the new ones we are adding which is clean to look at in a diff, but would it be nicer to sort the whole list so that anyone looking at the code later on sees an ordered list rather than random stuff added over time?

esword (commented):
pretty sweet when you get something an order of magnitude better for 1/2 the code.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/CodeChangesTest.groovy
```diff
@@ -0,0 +1,217 @@
...
+    // Change this to be the filename of the test (copy from the file viewer) to run a single test
+    // static String testOverride = 'AugmentExistingSuppresionsList.java'
+    static String testOverride = ''
+
+    @Unroll
+    def '#testFile'() {
```
esword: why have this custom format of the before and after in a single file vs having before and expected versions of the file?  This seems pretty fragile and it also makes it harder to see what the actual changes are - if they are in 2 different files a reader can just diff them in them in the IDE to see what the intended state is.

esword: or for the methods-type just doing it in the test case/file since it's a bunch of short text.

CRogers: I've come to really not like having test inputs and outputs be in different files, I find it makes it harder to see what the test should do as you have to line and flick between the files. It's clearer (for me at least) when they're right next to each other. Additionally, in these tests each file requires running gradle so it's best to group some together so it doesn't take loads of time to run.

Your IDE already diffs for you when it goes wrong, no need to file based diff:

Screenshot 2025-01-16 at 20 05 13

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
CRogers: It actually _can't_ be a memoized supplier - it took me ages to debug when I initially made it like that. I'll add a comment.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -63,60 +75,76 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
     }
 
-    static CharSequence whitespaceIndentBefore(CharSequence sourceCode, int sourceElementPosition) {
-        int pos = sourceElementPosition - 1;
+    private static Optional<ModifiersTree> isModifiersTree(Tree tree) {
+        // This covers all type definitions eg class, interface, enum, record, annotation, future kinds
+        // of class-like type definitions.
+        if (tree instanceof ClassTree) {
+            return Optional.of(((ClassTree) tree).getModifiers());
+        }
```
felixdesouza, line 126: am I misremembering or was there a reason you couldn't use java 17 here?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
felixdesouza: +1

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -132,7 +129,7 @@ private static void setupErrorProneArtifactTransform(Project project) {
                 .attribute(suppressible, false);
 
         project.getDependencies().registerTransform(ModifyErrorProneCheckApi.class, spec -> {
-            spec.getParameters().getSuppressionStage1().set(isSuppressingStageOne(project));
+            spec.getParameters().getSuppressionStage1().set(isSuppressing(project));
```
felixdesouza: nit: `getSuppressionStage1` -> `getIsSuppressing()`

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -17,25 +17,37 @@
...
 import java.util.logging.Logger;
 import java.util.stream.Stream;
+import javax.lang.model.element.Name;
 // CHECKSTYLE:ON
 
 public final class VisitorStateModifications {
     private static final Logger log = Logger.getLogger(VisitorStateModifications.class.getName());
 
+    // Weak map so that we don't leak memory by keeping our mutable fixes around forever, once error-prone has done
+    // with them, they can be safely collected.
```
felixdesouza:
```suggestion
    // Weak map so that we don't leak memory by keeping our mutable fixes around forever, once errorprone
    // finished with a class, our SuppressingFix's can be safely collected.
```

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/CodeChangesTest.groovy
felixdesouza: I agree with @esword making a new format is a bit meh, similarly agree with @CRogers that diffing across files isn't amazing. I suppose theres the "third" thing of having to go to other files is kinda meh as well.

I actually just prefer to see it in the test file itself 😅 

I'd be curious if you could do something like:

```groovy
class CodeChangesTest {

    List<TestCase> testCases = List.of(testCase(
            name: 'AddNewSuppressionsAtEveryLevel',
            args: '-PerrorProneSuppress',
            //language=Java
            before: '''
                public final class App {
                    public final String field = new int[3].toString();
                
                    public App() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void method() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void variables() {
                        String variable = new int[3].toString();
                        System.out.println(variable);
                    }
                
                    public static class SomeClass {
                        static {
                            System.out.println(new int[3].toString());
                        }
                    }
                }
            ''',
            //language=Java
            after: '''
              public final class App {
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public final String field = new int[3].toString();
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public App() {
                      System.out.println(new int[3].toString());
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public void method() {
                      System.out.println(new int[3].toString());
                  }
              
                  public void variables() {
                      @SuppressWarnings("for-rollout:ArrayToString")
                      String variable = new int[3].toString();
                      System.out.println(variable);
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public static class SomeClass {
                      static {
                          System.out.println(new int[3].toString());
                      }
                  }
              }  
            '''
    ))

    def '#testName bla'() {
        setup:
        // bla

        when:
        // bla

        then:
        // bla

        where:
        testCase << testCases
        name = testCase.name
        args = testCase.args
        before = testCase.before
        after = testCase.after
    }
}
```

felixdesouza: or you could do another thing where you have methods that you reflect into and call e.g.

```groovy
@TestCase(args = '-PerrorProneSuppress')
TestCase 'AddNewSuppressionsAtEveryLevel'() {
  testCase(
    //language=Java
            before: '''
                public final class App {
                    public final String field = new int[3].toString();
                
                    public App() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void method() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void variables() {
                        String variable = new int[3].toString();
                        System.out.println(variable);
                    }
                
                    public static class SomeClass {
                        static {
                            System.out.println(new int[3].toString());
                        }
                    }
                }
            ''',
            //language=Java
            after: '''
              public final class App {
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public final String field = new int[3].toString();
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public App() {
                      System.out.println(new int[3].toString());
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public void method() {
                      System.out.println(new int[3].toString());
                  }
              
                  public void variables() {
                      @SuppressWarnings("for-rollout:ArrayToString")
                      String variable = new int[3].toString();
                      System.out.println(variable);
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public static class SomeClass {
                      static {
                          System.out.println(new int[3].toString());
                      }
                  }
              }  
            '''
    )
}
```

felixdesouza: you get all the benefits you describe, but you don't have to go to a different file to find the source and what you expect etc.

you also have a lot less parsing logic to do and can just directly encode it

felixdesouza: groovy also gives you ability to have a nice dsl instead, so you could do:

```groovy
@TestCase()
TestCase 'AddNewSuppressionsAtEveryLevel'() {
  testCase {
            args '-PerrorProneSuppress'
            //language=Java
            before '''
                public final class App {
                    public final String field = new int[3].toString();
                
                    public App() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void method() {
                        System.out.println(new int[3].toString());
                    }
                
                    public void variables() {
                        String variable = new int[3].toString();
                        System.out.println(variable);
                    }
                
                    public static class SomeClass {
                        static {
                            System.out.println(new int[3].toString());
                        }
                    }
                }
            '''
            
            //language=Java
            after '''
              public final class App {
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public final String field = new int[3].toString();
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public App() {
                      System.out.println(new int[3].toString());
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public void method() {
                      System.out.println(new int[3].toString());
                  }
              
                  public void variables() {
                      @SuppressWarnings("for-rollout:ArrayToString")
                      String variable = new int[3].toString();
                      System.out.println(variable);
                  }
              
                  @SuppressWarnings("for-rollout:ArrayToString")
                  public static class SomeClass {
                      static {
                          System.out.println(new int[3].toString());
                      }
                  }
              }  
            '''
    }
}
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: Some repos internally run the compiler with 11, we had to revert back in #22

CRogers: Actually, thanks to https://github.com/palantir/gradle-baseline/pull/2898, baseline 6.10.0 now ships java 17 errorprones, so these repos are more directly blocked from there.

CRogers: Looks like 46 repos have been stuck on baseline 6.9.0 because of that change

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
CRogers: I like it, let's keep the for-rollout ones sorted and at the end.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/CodeChangesTest.groovy
CRogers: Ok, I've reverted back to something way less silly

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtilsTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        @Test
+        void maintains_alphabetical_order_for_automated_suppressions() {
+            assertThat(SuppressWarningsUtils.modifySuppressions(
+                            List.of("Blah", "for-rollout:A", "for-rollout:Something"), Set.of("ArrayToString")))
+                    .containsExactly("Blah", "for-rollout:A", "for-rollout:ArrayToString", "for-rollout:Something");
+        }
+
+        @Test
+        void reorders_automated_suppresions_to_the_end() {
```
felixdesouza, line 61: I think you want another test where it's:

``java
assertThat(SuppressWarningsUtils.modifySuppressions(
                            List.of("Blah", "for-rollout:A", "HumanAuthored"), Set.of("ArrayToString")))
```

where you end up with `"Blah", "HumanAuthored", "for-rollout:A", "for-rollout:ArrayToString"`

this is for the case where someone has added another suppression after some started rolling out

File: suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtilsTest.java [hunk 14]
```diff
@@ -0,0 +1,93 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class SuppressWarningsUtilsTest {
+    @Nested
```
felixdesouza, line 27: neat

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtilsTest.java
CRogers: Done

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtils.java
```diff
@@ -0,0 +1,78 @@
...
+                automaticallyAddedOrNotSuppressions.getOrDefault(SuppressionsType.HUMAN_AUTHORED, List.of());
+
+        List<String> existingAutomaticallyAddedSuppressionsWithoutPrefix =
+                existingAutomaticallyAddedSuppressions.stream()
+                        .map(error -> error.replace(CommonConstants.AUTOMATICALLY_ADDED_PREFIX, ""))
+                        .collect(Collectors.toList());
+
+        List<String> modifiedAutomaticallyAddedSuppressions = Stream.concat(
+                        existingAutomaticallyAddedSuppressionsWithoutPrefix.stream(), newAutomatedSuppressions.stream())
+                .filter(Predicate.not(humanAuthoredSuppressions::contains))
```
felixdesouza, line 54: I wish java would just take `append` and `remove` from `StreamEx`, it would become this:

```java
        List<String> modifiedAutomaticallyAddedSuppressions = StreamEx
                .of(existingAutomaticallyAddedSuppressionsWithoutPrefix)
                .append(newAutomatedSuppressions)
                .remove(humanAuthoredSuppressions::contains)
```

felixdesouza: I take it `StreamEx` is not a dependency you've taken here

CRogers: +1,000,000

felixdesouza: not blocking, just sad 😂

CRogers: I just want to minimise deps, even if the code is a little more verbose

felixdesouza:
👍  let's goooo

PR #40, opened 2025-01-23
Fix error where suppressing could produce incorrect `App@SuppressWarnings("Check").Builder` expressions

CRogers:
## Before this PR
We were producing expressions of the form:

```java
App@SuppressWarnings("Check").Builder variable = <expression-with-error>;
```

instead of the correct:

```java
@SuppressWarnings("Check")
App.Builder variable = <expression-with-error>;
```

This is because of the new approach in #39, we hold onto source elements for much longer, and evaluate source code positions of tree elements in later compiler passes. The start positions of these tree elements have then changed, resulting in the bug above. There are detailed comments in the diff.

## After this PR

==COMMIT_MSG==
Fix error where suppressing could produce incorrect `App@SuppressWarnings("Check").Builder` expressions
==COMMIT_MSG==

We fix the problem by evaluating this information as soon as we can, rather than deferring until error-prone actually does the replacement applications.

## Possible downsides?
The fact that we have to evaluate different bits of information for the replacement at different stages (replacement range earlier, replacement string later), as well as getting the `EndPosTable` at a different to do this, makes this more complicated. Now there is code that is more comments than code to explain it :)

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -451,6 +451,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         writeJavaSourceFileToSourceSets '''
             package app;
             public final class App {
+                @SuppressWarnings("UnusedVariable")
                 public static void main(String[] args) {
                     App.Builder builder = new App.Builder(new int[3].toString());
                 }
@@ -468,6 +469,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         appJavaTextEquals '''
             package app;
             public final class App {
+                @SuppressWarnings("UnusedVariable")
                 public static void main(String[] args) {
                     @SuppressWarnings("for-rollout:ArrayToString")
                     App.Builder builder = new App.Builder(new int[3].toString());
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -301,9 +301,10 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         runTasksSuccessfully('compileAllErrorProne', '-PerrorProneSuppress')
 
         then:
-        runTasksSuccessfully('compileAllErrorProne')
-
         appJavaTextContains('@SuppressWarnings(\"for-rollout:ArrayToString\")')
+
+
+        runTasksSuccessfully('compileAllErrorProne')
     }
 
     def 'demonstrate suppressions on different source elements'() {
@@ -338,8 +339,6 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         runTasksSuccessfully('compileAllErrorProne', '-PerrorProneSuppress')
 
         then:
-        runTasksSuccessfully('compileAllErrorProne')
-
         // language=Java
         appJavaTextEquals '''
             package app;
@@ -371,6 +370,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
                 }
             }
         '''.stripIndent(true)
+
+        runTasksSuccessfully('compileAllErrorProne')
     }
 
     def 'supports errorprone checks that match on a larger element than they report errors on'() {
@@ -393,8 +394,6 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         runTasksSuccessfully('compileAllErrorProne', '-PerrorProneSuppress')
 
         then:
-        runTasksSuccessfully('compileAllErrorProne')
-
         // language=Java
         appJavaTextEquals '''
             package app;
@@ -405,6 +404,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
                 }
             }
         '''.stripIndent(true)
+
+        runTasksSuccessfully('compileAllErrorProne')
     }
 
     def 'supports suppressing errorprone checks on classes, interfaces, records, enums, etc'() {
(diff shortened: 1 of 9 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -301,9 +301,10 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         runTasksSuccessfully('compileAllErrorProne', '-PerrorProneSuppress')
 
         then:
-        runTasksSuccessfully('compileAllErrorProne')
-
         appJavaTextContains('@SuppressWarnings(\"for-rollout:ArrayToString\")')
+
+
+        runTasksSuccessfully('compileAllErrorProne')
```
CRogers, line 307: The error messages are better if you just get an IDE powered diff right away rather than "compilation failed, scroll up through output"

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AnnotationUtils.java
```diff
@@ -0,0 +1,55 @@
...
+            if (!(arg instanceof AssignmentTree)) {
+                return Stream.empty();
+            }
+            AssignmentTree assignment = (AssignmentTree) arg;
+
+            ExpressionTree expression = assignment.getExpression();
+
+            if (expression instanceof LiteralTree) {
+                LiteralTree literalTree = (LiteralTree) expression;
+                return Stream.of((String) literalTree.getValue());
```
felixdesouza, line 38: is this a safe cast?

CRogers: It is if the only literal values the annotations has are string (like `@SuppressWarnings` does)

CRogers: This code was here before, but perhaps something could go wrong if someone did:

```java
static String CHECK_NAME = "blah";

@SuppressWarnings(CHECK_NAME)
```

CRogers: Internally, we have zero usages of constants in suppress warnings, but we do have this `value` variant in a few hundred places:

```java
@SuppressWarnings(value = {"NullAway", ...})
```

which I can make sure we support in another PR.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
```diff
@@ -0,0 +1,101 @@
...
+    private final Set<String> newSuppressions;
+
+    SuppressingReplacement(
+            EndPosTable endPositions,
+            Set<String> newSuppressions,
+            Optional<CharSequence> sourceCode,
+            Optional<? extends AnnotationTree> suppressWarnings,
+            Tree tree) {
+        this.newSuppressions = newSuppressions;
+        this.range = calculateRange(endPositions, suppressWarnings, tree);
```
felixdesouza: perhaps move the bit about calculating the range eagerly just above this line?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
```diff
@@ -17,120 +17,40 @@
...
-    public void suppressError(String error) {
-        errors.add(error);
+    public void addSuppression(String suppression) {
+        newSuppressions.add(suppression);
     }
 
     @Override
     public ImmutableSet<Replacement> getReplacements(EndPosTable endPositions) {
-        return ImmutableSet.of(
-                new SuppressingReplacement(() -> Iterables.getOnlyElement(fix().getReplacements(endPositions))));
```
felixdesouza, line 62: so if I've understood correctly, `getReplacements` can be called multiple times and in effect you'll keep making thunks?

I suppose why not just use a `ConcurrentMap`? or just `Map`? with a `computeIfAbsent` method?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
```diff
@@ -0,0 +1,101 @@
...
+
+        this.existingSuppressions = suppressWarnings
+                .map(AnnotationUtils::annotationStringValues)
+                .orElseGet(Stream::of)
```
felixdesouza: nit: `Stream::empty` or

```java
suppressWarnings.stream()
                .flatMap(AnnotationUtils::annotationStringValues)
                .collect(Collectors.toList());
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java
felixdesouza: If it's not the above, I think I'm not fully understanding the subtlety. From what you described in person, I'm not sure how this particular chunk in this diff makes it work 😬

CRogers: I thought about it, but then I have to save all the `sourceCode`, `tree`, `suppressWarnings` as fields rather than just consuming them in the constructor, which I think adds a bit more clutter (could hide it behind a Function - but that also feels a bit much and I'd end up wrapping it up in a type - which is basically `FirstTimeMemoizingFunction`). Once you've made the type feels a bit wasteful to make a whole map when there's just one value.

CRogers: > `getReplacements` can be called multiple times and in effect you'll keep making thunks?

`getReplacements` gets called multiple times, but only ever have one thunk/`SuppressingReplacement` thanks to the `FirstTimeMemoizingFunction`

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
CRogers: done

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
CRogers: done

felixdesouza (commented):
👍

PR #43, opened 2025-01-30
Allow running `-PerrorProneApply` and `-PerrorProneSuppress` at the same time

CRogers:
## Before this PR
Currently, when rolling out errorprone changes, we first apply suggested fixes then suppress anything that was not autofixed. This requires two whole compilations:

```
./gradlew compileAllErrorProne -PerrorProneApply
./gradlew compileAllErrorProne -PerrorProneSuppress
```

This is quite wasteful, especially as some internal projects take nearly 10 mins to compile, this is can get very close to the default excavator timeout when run twice.

I also want to get nit-bot to suppress errorprone errors introduced in excavator PRs, which is quite a problem. For example, caffeine introduced a change that angers `NullAway`, and there are many failing excavators that have not worked because this was not suppressed. It would be very wasteful to run errorprone suppression on every excavator, hence nit-bot is ideal to detect when it is needed. Currently, nit-bot only allows one command (I guess I could use a shell to run two runs...) but makes sense to avoid wasting nit-bot execution time.

## After this PR
==COMMIT_MSG==
Running `-PerrorProneApply` and `-PerrorProneSuppress` at the same time is now possible, saving compilation time.
==COMMIT_MSG==

## Possible downsides?
Slightly more complexity, there is another "mode".

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -37,7 +37,7 @@ import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.process.CommandLineArgumentProvider;
 
 public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
-    private static final String SUPPRESS_STAGE = "errorProneSuppress";
+    private static final String ERROR_PRONE_SUPPRESS = "errorProneSuppress";
     private static final String ERROR_PRONE_APPLY = "errorProneApply";
     private static final String ERROR_PRONE_DISABLE = "errorProneDisable";
     private static final String ERROR_PRONE_TIMINGS = "errorProneTimings";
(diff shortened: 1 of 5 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -253,6 +252,28 @@ public Iterable<String> asArguments() {
         }
     }
 
+    private static List<String> checksToApplySuggestedPatchesFor(
```
felixdesouza, line 255: curious, when does this run? does the output of this feed into up-to-date checks? I'm seeing the argument provider and not sure when things are called etc.

CRogers: It's used in the two of the argument providers above. Basically any time `-PerrorProneApply` is set, but if `-PerrorProneSuppress` is also set we need to do something different, so extracted to a method.

Also, re: up-to-date, these compilations rerun every time basically.

felixdesouza (approved):
🚀

PR #57, opened 2025-03-17
Support auto-removing for-rollout suppression warnings

aldexis:
## Before this PR
This plugin adds `for-rollout:` suppressions for the error-prone checks, but doesn't have a way to get rid of them in a simple and automated manner.

This is useful for cases such as:
- Get a new check rolled out, suppressed on existing usage
- Automatically open a new pull request with the suppressions removed, to make it clear where they need to be removed

Specifically, we should only remove the for-rollout suppressions, and not the manual ones.

## After this PR
==COMMIT_MSG==
Support auto-removing for-rollout suppression warnings
==COMMIT_MSG==

Note that this does _not_ distinguish between errors that are still active vs not. It simply removes all instances of the suppressions.

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -38,6 +38,7 @@ import org.gradle.process.CommandLineArgumentProvider;
 
 public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
     private static final String ERROR_PRONE_SUPPRESS = "errorProneSuppress";
+    private static final String ERROR_PRONE_REMOVE_SUPPRESSIONS = "errorProneRemove";
     private static final String ERROR_PRONE_APPLY = "errorProneApply";
     private static final String ERROR_PRONE_DISABLE = "errorProneDisable";
     private static final String ERROR_PRONE_TIMINGS = "errorProneTimings";
@@ -54,8 +55,13 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
 
     private void applyToJavaProject(Project project) {
         if (isDisabled(project) && isAnyKindOfPatching(project)) {
+            throw new IllegalStateException("-PerrorProneDisable cannot be used at the same time as "
+                    + "-PerrorProneApply, -PerrorProneSuppress or -PerrorProneRemove");
+        }
+
+        if (isRemovingSuppressions(project) && (isSuppressing(project) || isApplyingSuggestedPatches(project))) {
             throw new IllegalStateException(
-                    "-PerrorProneDisable cannot be used at the same time as -PerrorProneApply or -PerrorProneSuppress");
+                    "-PerrorProneRemove cannot be used at the same time as -PerrorProneSuppress or -PerrorProneApply");
         }
 
         project.getPluginManager().apply(ErrorPronePlugin.class);
@@ -203,6 +209,28 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
 
         errorProneOptions.getExcludedPaths().set(excludedPathsRegex());
 
+        if (isRemovingSuppressions(project)) {
+            errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
+                @Override
+                public Iterable<String> asArguments() {
+                    String argument = (String) javaCompile.getProject().property(ERROR_PRONE_REMOVE_SUPPRESSIONS);
+
+                    List<String> suppressionsToRemove = Arrays.stream(argument.split(","))
+                            .map(String::trim)
+                            .filter(Predicate.not(String::isEmpty))
+                            .toList();
+
+                    return List.of(
+                            "-XepPatchLocation:IN_PLACE",
+                            "-XepPatchChecks:RemoveSuppressions",
+                            "-XepOpt:SuppressibleErrorProne:RemoveForRolloutWarnings="
+                                    + String.join(",", suppressionsToRemove));
+                }
+            });
+
+            return;
+        }
+
         if (isSuppressing(project)) {
             errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
                 @Override
@@ -285,7 +313,7 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
     }
 
     private static boolean isAnyKindOfPatching(Project project) {
-        return isApplyingSuggestedPatches(project) || isSuppressing(project);
+        return isApplyingSuggestedPatches(project) || isSuppressing(project) || isRemovingSuppressions(project);
     }
 
     private static boolean isApplyingSuggestedPatches(Project project) {
@@ -294,6 +322,10 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
 
     private static boolean isSuppressing(Project project) {
         return project.hasProperty(SuppressibleErrorPronePlugin.ERROR_PRONE_SUPPRESS);
+    }
+
+    private static boolean isRemovingSuppressions(Project project) {
+        return project.hasProperty(SuppressibleErrorPronePlugin.ERROR_PRONE_REMOVE_SUPPRESSIONS);
     }
 
     static String excludedPathsRegex() {
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -56,12 +56,12 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
     private void applyToJavaProject(Project project) {
         if (isDisabled(project) && isAnyKindOfPatching(project)) {
             throw new IllegalStateException("-PerrorProneDisable cannot be used at the same time as "
-                    + "-PerrorProneApply, -PerrorProneSuppress or -PerrorProneRemove");
+                    + "-PerrorProneApply, -PerrorProneSuppress or -PerrorProneRemoveRollout");
         }
 
         if (isRemovingSuppressions(project) && (isSuppressing(project) || isApplyingSuggestedPatches(project))) {
-            throw new IllegalStateException(
-                    "-PerrorProneRemove cannot be used at the same time as -PerrorProneSuppress or -PerrorProneApply");
+            throw new IllegalStateException("-PerrorProneRemoveRollout cannot be used at the same time as "
+                    + "-PerrorProneApply or -PerrorProneSuppress");
         }
 
         project.getPluginManager().apply(ErrorPronePlugin.class);
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -213,7 +213,7 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
             errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
                 @Override
                 public Iterable<String> asArguments() {
-                    List<String> suppressionsToRemove = checksToRemoveSuppressionsFor(javaCompile);
+                    List<String> suppressionsToRemove = checksToRemoveSuppressionsFor(project);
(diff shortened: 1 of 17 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -54,8 +55,13 @@ public void apply(Project project) {
...
         if (isDisabled(project) && isAnyKindOfPatching(project)) {
-            throw new IllegalStateException(
-                    "-PerrorProneDisable cannot be used at the same time as -PerrorProneApply or -PerrorProneSuppress");
+            throw new IllegalStateException("-PerrorProneDisable cannot be used at the same time as "
+                    + "-PerrorProneApply, -PerrorProneSuppress or -PerrorProneRemoveRollout");
+        }
+
+        if (isRemovingSuppressions(project) && (isSuppressing(project) || isApplyingSuggestedPatches(project))) {
+            throw new IllegalStateException("-PerrorProneRemoveRollout cannot be used at the same time as "
+                    + "-PerrorProneApply or -PerrorProneSuppress");
```
aldexis, line 64: I could potentially try to make this possible, but this would lead to more complexity in the error prone options setup and it doesn't seem like something that is needed for now, hence why I didn't

suppressible-error-prone/build.gradle
```diff
@@ -24,10 +24,8 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.util')
 }
 
-// Unfortunately, a few internal products are still compiling with Java 11. Remove this entirely once this is no
-// longer the case.
 javaVersion {
-    target = 11
+    target = 17
```
aldexis, line 28: Per discussion with @CRogers, this shouldn't be needed anymore

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AnnotationUtils.java
```diff
@@ -51,5 +55,18 @@ static Stream<String> annotationStringValues(AnnotationTree annotation) {
         });
     }
 
+    static Name annotationName(Tree annotationType) {
```
aldexis, line 58: Moved over from VisitorStateModifications

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -0,0 +1,162 @@
...
+     * Here we want to:
+     *   - replace the element defined by the provided position
+     *   - also replace the whitespace before the element, up to and including the newline
+     * This way, if e.g. @SuppressWarnings("foo") must be removed entirely, we can remove the entire line, rather than
+     *   just the annotation, leaving us with an empty line.
+     *
+     * Note that this will only delete the whitespace before the element if the entire element is removed
+     *   (i.e. if the replacement text is null or empty).
+     */
+    private static final class LineRemovingReplacementFix implements Fix {
```
aldexis, line 107: Unfortunately, I didn't find a good way to remove the entire line/whitespace before the tree we want to replace, so I introduced my own version of it, similar to SuppressingFix

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -0,0 +1,162 @@
...
+            return CoalescePolicy.REJECT;
+        }
+
+        @Override
+        public ImmutableSet<Replacement> getReplacements(EndPosTable endPositions) {
+            if (replacementText.isEmpty() && sourceCode != null) {
+                int start = SourceCodeUtils.startPositionWithWhitespaceIncludingNewLine(
+                        sourceCode, position.getStartPosition());
+                return ImmutableSet.of(Replacement.create(start, position.getEndPosition(endPositions), ""));
+            }
```
aldexis, line 143: This is where the specific logic happens

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: Yeah, makes sense to just keep it simple and make it error out. Unlikely to both be applying/suppressing and removing _a specific_ errorprone suppression at the same time.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -276,6 +299,22 @@ private static List<String> checksToApplySuggestedPatchesFor(
                 .collect(Collectors.toList());
     }
 
+    private static List<String> checksToRemoveSuppressionsFor(JavaCompile javaCompile) {
```
CRogers: I think this can take `Project` instead of `JavaCompile`?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -276,6 +299,22 @@ private static List<String> checksToApplySuggestedPatchesFor(
...
+        String possibleChecksToRemove = (String) javaCompile.getProject().property(ERROR_PRONE_REMOVE_SUPPRESSIONS);
+
+        // For the suppressions to remove, if no specific check is enabled, we need to just remove everything
+        // We can't explicitly list all possible checks, because some might not exist anymore
+        // The logic itself needs to consider an empty list as "remove all"
+        if (possibleChecksToRemove == null) {
+            return List.of();
+        }
```
CRogers, line 310: Hmm, didn't think about this case. I guess it would make sense to do this? Maybe you do want to remove all automated suppressions at some point (I guess it would help with manually testing this plugin on a repo that already has everything suppressed)

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressionsTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+package com.palantir.suppressibleerrorprone;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import org.junit.jupiter.api.Test;
+
+class RemoveRolloutSuppressionsTest {
+
+    @Test
+    void testRemoveAllForRollout() {
```
CRogers: `test` prefixes on unit tested code is not necessary and just add visual noise, we also prefer to use snake_case for test names and write them as if they were an english sentence (eg `no_specific_checks_means_rollout_suppressions_are_removed` or something similar (tbh `remove_all_for_rollout` would probably also do)).

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressionsTest.java
```diff
@@ -0,0 +1,238 @@
...
+        fix().addInputLines(
+                        "App.java",
+                        // language=Java
+                        """
+                        package app;
```
CRogers: I don't think you need a package, you can just use the default package. Makes the tests a bit more concise/easier to read.

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressionsTest.java
aldexis: Oh good point - will update 👍

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
aldexis: At minimum, it makes sense to me to support it. It would also enable us to automatically open up PRs on repos to flag all existing for-rollout suppressions

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
aldexis: Indeed - didn't catch that javaCompile.getProject was essentially just the project 🤦

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressionsTest.java
```diff
@@ -0,0 +1,238 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.suppressibleerrorprone;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import org.junit.jupiter.api.Test;
+
+class RemoveRolloutSuppressionsTest {
```
CRogers, line 22: Might be worth throwing in a single test that checks it does stuff on other element types eg fields, variables, methods etc - this test here demonstrates all of them:

https://github.com/palantir/suppressible-error-prone/blob/35154da4616375ce22522f64d79da99c339d23db/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy#L348-L369

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/WhitespaceIndentBeforeTest.java
```diff
@@ -43,6 +43,28 @@ void something_else_is_before_on_the_same_line() {
         assertThat(whitespaceIndentBefore("class Foo {} |class Bar {}")).isEqualTo(" ");
     }
 
+    @Test
```
CRogers: I think it's nice to use junit5's `@Nested` here to group together tests for entirely different methods - also means you don't need to prefix the test names with "start position". [Example](https://github.com/palantir/gradle-guide/blob/a626b8e1cfeddcd1c34decde5712e4955adae2a0/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistrationTest.java#L109)

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/WhitespaceIndentBeforeTest.java
```diff
@@ -43,6 +43,28 @@ void something_else_is_before_on_the_same_line() {
...
+    void startPositionWithNewline() {
```
CRogers: Would prefer this to be snake_case for test names

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -0,0 +1,164 @@
...
+        private final CharSequence sourceCode;
+        private final DiagnosticPosition position;
+        private final String replacementText;
+
+        private LineRemovingReplacementFix(
+                CharSequence sourceCode, DiagnosticPosition position, String replacementText) {
+            this.sourceCode = sourceCode;
+            this.position = position;
+            // Guarantee replacementText isn't empty to simplify the checks below
+            this.replacementText = replacementText == null ? "" : replacementText;
```
CRogers: Would `replacementText` ever be null?

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressionsTest.java
aldexis: Oh yeah, I had wanted to do this and forgot. Thanks for flagging!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
aldexis: Not in our case actually, so I can simplify the code here indeed.

suppressible-error-prone/src/test/java/com/palantir/suppressibleerrorprone/WhitespaceIndentBeforeTest.java
aldexis: I didn't know about it - this is pretty nice 👍

CRogers:
:+1:

PR #67, opened 2025-04-07
Avoid `Note: [RemoveRolloutSuppressions]` in unrelated error messages

CRogers:
## Before this PR
People are getting confused by `Note: [RemoveRolloutSuppressions]` appearing in unrelated error messages, thinking this was the actual error message:

```
Compilation failed; see the compiler output below.
src/main/java/app/App.java:4: Note: [RemoveRolloutSuppressions] Remove for-rollout suppression warnings
    @SuppressWarnings("for-rollout:NullAway")
    ^
    (see https://github.com/palantir/suppressible-error-prone)
  Did you mean 'public final class App {'?
src/main/java/app/App.java:6: error: [ArrayToString] Calling toString on an array does not provide useful information
        System.out.println(new int[3].toString());
                           ^
    (see https://errorprone.info/bugpattern/ArrayToString)
  Did you mean 'System.out.println(Arrays.toString(new int[3]));'?
1 error
```

This is because we have `RemoveRolloutSuppressions` as an enabled Suggestion level check which we use to remove rollout suppressions.

## After this PR
==COMMIT_MSG==
Avoid `Note: [RemoveRolloutSuppressions]` in unrelated error messages
==COMMIT_MSG==

We just disable the check now if we don't need it, which stops these issues.

### DIFF
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
@@ -51,7 +51,7 @@ import javax.lang.model.element.Name;
         suppressionAnnotations = {})
 public final class RemoveRolloutSuppressions extends BugChecker implements BugChecker.AnnotationTreeMatcher {
 
-    public static final String ARGUMENT = "SuppressibleErrorProne:RemoveForRolloutWarnings";
+    public static final String ARGUMENT = "SuppressibleErrorProne:RemoveRolloutSuppressions";
 
     @Override
     public Description matchAnnotation(AnnotationTree tree, VisitorState state) {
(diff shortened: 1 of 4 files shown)


suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -51,7 +51,7 @@
         suppressionAnnotations = {})
 public final class RemoveRolloutSuppressions extends BugChecker implements BugChecker.AnnotationTreeMatcher {
 
-    public static final String ARGUMENT = "SuppressibleErrorProne:RemoveForRolloutWarnings";
```
CRogers, line 54: This annoyed me, since it's all internal did a quick fix

bmarcaur (approved):

felixdesouza (approved):

PR #69, opened 2025-04-09
Allow patching specific unsuppressed checks when using errorProneRemoveRollout

aldexis:
## Before this PR
When removing for-rollout suppressions, we aren't actually applying fixes that we might (e.g. [SafeLoggingPropagation](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java)).

This leads to more friction than necessary, as these will now fail the compilation, but could be automatically fixed.

## After this PR
==COMMIT_MSG==
Allow patching specific unsuppressed checks when using errorProneRemoveRollout
==COMMIT_MSG==

## Possible downsides?
- ~It might be confusing that we're automatically fixing these without using a dedicated argument~
- It might be confusing if pull requests are being opened that typically require manual intervention (and mention it as such in their body), but instead are automatically fixed and merged.

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -19,8 +19,10 @@ package com.palantir.gradle.suppressibleerrorprone;
 import com.palantir.gradle.suppressibleerrorprone.transform.ModifyErrorProneCheckApi;
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import net.ltgt.gradle.errorprone.CheckSeverity;
@@ -75,15 +77,18 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
                 .orElseThrow(
                         () -> new RuntimeException("SuppressibleErrorPronePlugin implementation version not found"));
 
-        // When auto-suppressing, there are two stages:
-        // 1. The first runs a bytecode patched version of errorprone (via an
-        //    artifact transform) that intercepts every error from every check and adds a custom fix, a
-        //    @RepeatableSuppressWarnings annotation to the relevant statement/method/field/class.
-        // 2. The second stage runs a single errorprone check: SuppressWarningsCoalesce, which will combine all
-        //    the @RepeatableSuppressWarnings and @SuppressWarnings annotations into one normal @SuppressWarnings
-        //    annotation.
+        // Don't need to consider the custom suppression logic if we're going to remove the suppressions anyway
+        if (!isRemovingSuppressions(project)) {
+            // When auto-suppressing, there are two stages:
+            // 1. The first runs a bytecode patched version of errorprone (via an
+            //    artifact transform) that intercepts every error from every check and adds a custom fix, a
+            //    @RepeatableSuppressWarnings annotation to the relevant statement/method/field/class.
+            // 2. The second stage runs a single errorprone check: SuppressWarningsCoalesce, which will combine all
+            //    the @RepeatableSuppressWarnings and @SuppressWarnings annotations into one normal @SuppressWarnings
+            //    annotation.
 
-        setupErrorProneArtifactTransform(project);
+            setupErrorProneArtifactTransform(project);
+        }
 
         project.getConfigurations().named(ErrorPronePlugin.CONFIGURATION_NAME).configure(errorProneConfiguration -> {
             // Required so that we can run the runtime parts of the errorprone patching in suppressing stage 1 and
@@ -214,10 +219,14 @@ public final class SuppressibleErrorPronePlugin implements Plugin<Project> {
                 @Override
                 public Iterable<String> asArguments() {
                     List<String> suppressionsToRemove = checksToRemoveSuppressionsFor(project);
+                    Set<String> checksToPatch = new HashSet<>();
+                    checksToPatch.add("RemoveRolloutSuppressions");
+                    // TODO(aldexis): Should we only do this if a specific argument is present?
+                    checksToPatch.addAll(suppressionsToRemove);
 
                     return List.of(
                             "-XepPatchLocation:IN_PLACE",
-                            "-XepPatchChecks:RemoveRolloutSuppressions",
+                            "-XepPatchChecks:" + String.join(",", checksToPatch),
                             "-XepOpt:SuppressibleErrorProne:RemoveRolloutSuppressions="
                                     + String.join(",", suppressionsToRemove));
                 }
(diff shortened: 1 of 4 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -75,15 +77,18 @@ private void applyToJavaProject(Project project) {
...
+            // When auto-suppressing, there are two stages:
+            // 1. The first runs a bytecode patched version of errorprone (via an
+            //    artifact transform) that intercepts every error from every check and adds a custom fix, a
+            //    @RepeatableSuppressWarnings annotation to the relevant statement/method/field/class.
+            // 2. The second stage runs a single errorprone check: SuppressWarningsCoalesce, which will combine all
+            //    the @RepeatableSuppressWarnings and @SuppressWarnings annotations into one normal @SuppressWarnings
+            //    annotation.
+
+            setupErrorProneArtifactTransform(project);
+        }
```
aldexis: I _think_ this should be fine, but curious if you can think of problems this might cause

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -214,10 +219,14 @@ private void setupErrorProneOptions(
                 @Override
                 public Iterable<String> asArguments() {
                     List<String> suppressionsToRemove = checksToRemoveSuppressionsFor(project);
+                    Set<String> checksToPatch = new HashSet<>();
+                    checksToPatch.add("RemoveRolloutSuppressions");
+                    // TODO(aldexis): Should we only do this if a specific argument is present?
+                    checksToPatch.addAll(suppressionsToRemove);
```
aldexis: I'm wondering whether I should make a new argument and/or reuse errorProneApply (potentially with an extra check that you can't specify specific different checks to patch if doing the remove suppressions as well, since we should only patch these)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -75,15 +77,21 @@ private void applyToJavaProject(Project project) {
...
+        // Note that this means we need to handle the requested patches with care, so as to not apply patches to
+        //   checks that are suppressed with for-rollout, but for which we're not going to remove the suppressions.
+        if (!isRemovingSuppressions(project)) {
+            // When auto-suppressing, there are two stages:
+            // 1. The first runs a bytecode patched version of errorprone (via an
+            //    artifact transform) that intercepts every error from every check and adds a custom fix, a
+            //    @RepeatableSuppressWarnings annotation to the relevant statement/method/field/class.
+            // 2. The second stage runs a single errorprone check: SuppressWarningsCoalesce, which will combine all
+            //    the @RepeatableSuppressWarnings and @SuppressWarnings annotations into one normal @SuppressWarnings
+            //    annotation.
```
CRogers: I know this is unrelated, but this is actually wrong now, really it's just stage 1 and doesn't produce `@RepeatableSuppressWarnings` - I must have forgotten to update it.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -215,9 +223,27 @@ private void setupErrorProneOptions(
...
+                    // Otherwise, we might apply fixes for a check that we're not removing the suppression for, if
+                    //   suppressed using a for-rollout suppression (because we're not applying the for-rollout hack)
+                    if (isApplyingSuggestedPatches(project)) {
+                        Set<String> suppressionsToRemoveSet = new HashSet<>(suppressionsToRemove);
+                        List<String> extraChecksToPatch =
+                                checksToApplySuggestedPatchesFor(extension, javaCompile, errorProneOptions);
+                        if (extraChecksToPatch.stream().anyMatch(check -> !suppressionsToRemoveSet.contains(check))) {
+                            throw new IllegalStateException(
+                                    "Checks to patch must be a subset of the checks to remove suppressions for");
+                        }
```
CRogers: I'd actually collect the checks that aren't in the remove set to a list and check it's empty. Then we can paste all the ones that are not there in the error message (I don't really like error messages that don't show this kind of info).

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
aldexis: Should I remove the entire comment? I don't think I understand enough of the internals to provide a properly updated comment. Proposal would be
```
When auto-suppressing, the logic will run a bytecode patched version of errorprone (via an artifact transform) that intercepts every error from every check and adds a custom fix
```
but I'm not sure that's correct

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
aldexis: Yeah, fair enough!

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -976,6 +976,172 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
...
+        appJavaTextEquals '''
+            package app;
+            public final class App {
+                public static void main(String[] args) {
+                    System.out.println(new int[3].toString());
+                }
+            }
+        '''.stripIndent(true)
+    }
+
```
CRogers, line 1144: Might want a test for the `-PerrorProneRemove -PerrorProneApply` case, ie doing it for everything at once.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: yeah, that's good

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
aldexis: That specifically won't work, because `-PerrorProneRemove` doesn't create a list of all possible checks. It just accepts that an empty list should remove all `for-rollout` suppressions.

I guess I could however special-case this case if you think it's worth it. Where it might be a bit confusing though is that I would expect this to patch all checks that `-PerrorProneRemove` is removing suppressions for, which it won't unless the check is specifically configured to be patched (if I remember correctly how this works)

aldexis: I went ahead and special-cased this and added two more tests (one with specific patch checks, and one without)

CRogers (approved):

PR #70, opened 2025-04-10
[DO NOT MERGE] Reproduce bug where remove+apply doesn't work

aldexis:
## Before this PR
https://github.com/palantir/suppressible-error-prone/pull/69 introduced the possibility to remove the for-rollout suppressions at the same time as applying the fixes for said suppressions.

It turns out that this doesn't seem to work properly, if the fix is to apply another annotation.

## After this PR
==COMMIT_MSG==
Reproduce bug where remove+apply doesn't work
==COMMIT_MSG==

Currently, this PR only reproduces the bug, as I haven't been able to figure out how to fix it (tried all different CoalescePolicy and none seem to help)

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -1005,6 +1005,37 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         '''.stripIndent(true)
     }
 
+    def 'can patch checks while using -PerrorProneRemoveRollout, which also add annotations as fixes'() {
+        // language=Java
+        writeJavaSourceFileToSourceSets '''
+            package app;
+            public final class App {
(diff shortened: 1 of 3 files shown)


kelvinou01:
So I stepped thru the code for a bit during my work for the removeUnused mode. When removing rollout, we [take care to remove the whitespace and newlines around the suppression as well](https://github.com/palantir/suppressible-error-prone/blob/develop/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java#L142). This extends the start range of the fix to a few characters before the `@` in `@SuppressWarnings`, and potentially overlaps with other fixes like adding a `@Nullable`.

[In the case of overlapping fixes, error-prone will drop the incoming fix](https://github.com/google/error-prone/blob/a5a718974dd7d325025ea14c1492f113490d5cf8/check_api/src/main/java/com/google/errorprone/fixes/Replacements.java#L167). In this case, seems like the remove rollout fix is being dropped.

We could get around this by not extending the start range of the fix, and trusting the formatter to do the work. This successfully passes the bug reproducer test if you allow for some whitespace ugliness. A bunch of other tests also have to be loosened, whitespace and newline wise. We could run the formatter after compileAllErrorProne in the tests to get around this.

Another solution would be to intercept all fixes made on annotations, and consolidate them with our own fix. This way we ensure a suppressible tree (class/method/var) will only have at most one fix.

What do you think?

PR #95, opened 2025-06-11
Bump errorprone to 2.38.0

CRogers:
## Before this PR
Error-prone 2.38.0 has ABI breaks since 2.31.0. Internally, open source libraries that depend on `error-prone-annotations` are bumping up error-prone to 2.38.0, and these errors are manifesting. We should be on latest.

We upgraded baseline here: https://github.com/palantir/gradle-baseline/pull/3121

## After this PR
==COMMIT_MSG==
Bump errorprone to 2.38.0
==COMMIT_MSG==

Most of the test changes are just tidy up to get rid of the `System.out`s that errorprone now complains about.

## Possible downsides?
This repo has is stuck on a really old version of baseline, because the moment baseline started applying suppressible-error-prone, there is this weird self-referential error as Gradle tries to replace the `suppressible-error-prone` module with the local project (and requiring it's own jar before the :

```
> Could not resolve all files for configuration ':suppressible-error-prone:annotationProcessor'.
  > Failed to transform suppressible-error-prone-2.9.0-24-g6b3d671.dirty.jar (project :suppressible-error-prone) to match attributes {artifactType=jar, com.palantir.suppressible-error-prone.suppressible=true, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.version=17, org.gradle.libraryelements=jar, org.gradle.usage=java-runtime}.
    > Execution failed for ModifyErrorProneCheckApi: /Users/callumr/git/suppressible-error-prone/suppressible-error-prone/build/libs/suppressible-error-prone-2.9.0-24-g6b3d671.dirty.jar.
      > Transform output /Users/callumr/git/suppressible-error-prone/suppressible-error-prone/build/libs/suppressible-error-prone-2.9.0-24-g6b3d671.dirty.jar must exist.
```

To work around this, I've not upgraded the gradle plugin but have upgraded the `baseline-error-prone` jar that contains the error prone checks. We need this upgraded as these baseline checks were broken by upgrading error-prone.

I'm working on a long lived fix.


CRogers:
Superseded by #18

PR #97, opened 2025-06-24
Ensure all errorprone dependencies have the same version

CRogers:
## Before this PR
We have had two separate issues where error-prone dependencies in the compiler classpath (`annotationProcessor`) have been at different versions resulting in `ClassNotFoundError` or `NoSuchMethodError`s:

**First Issue**

```
Caused by: java.lang.ClassNotFoundException: com.google.errorprone.annotations.ImmutableTypeParameter
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
```

Caused by `error_prone_core` being at 2.38.0 while `error_prone_annotations` was forced down to 2.36.0 by a plugin's `enforcedPlatform`:

69c77179-ade6-4787-a4bc-7875110be920

**Second Issue**

```
     java.lang.NoSuchMethodError: 'com.google.errorprone.util.ErrorProneScope com.google.errorprone.util.ASTHelpers.scope(com.sun.tools.javac.code.Scope)'
        at com.google.errorprone.bugpatterns.ClassInitializationDeadlock.matchClass(ClassInitializationDeadlock.java:64)
```

This is actually a different issue where baseline is expecting error-prone 2.36.0 but it gets upgraded to 2.38.0. However, even here the versions of error-prone jars are different, which is likely to lead to issues:

Screenshot 2025-06-24 at 17 44 22

## After this PR
==COMMIT_MSG==
Ensure all errorprone dependencies have the same version
==COMMIT_MSG==

We do this using a `ComponentMetadataRule` and virtual platform, like GCV does for star dependencies. This is essentially the same as adding the following to `versions.props` for all consuming projects, but directly via Gradle:

```
com.google.errorprone:* = ...
```

For the First Issue above, this is the result - all the versions are the same:

Screenshot 2025-06-24 at 18 00 59

...and it says why (due to the platform):

Screenshot 2025-06-24 at 18 01 07
 

## Possible downsides?
I can't find a way to add a more specific reason for this (ie saying it comes from suppressible-error-prone) - but there is the platform explanation at least.

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -1226,6 +1226,34 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
         !rootCauseMessage.contains('[RemoveRolloutSuppressions]')
     }
 
+    def 'error-prone dependencies have versions bound together by a virtual platform'() {
+        setup: 'when an error-prone dependency is forced to certain version'
+        // language=Gradle
+        buildFile << '''
+            configurations.named('annotationProcessor') {
+                resolutionStrategy {
+                   force 'com.google.errorprone:error_prone_annotation:2.3.4'
+                }
+            }
+            
+            tasks.register('printErrorProneVersions') {
+                inputs.files(configurations.named('annotationProcessor'))
+                doLast {
+                    inputs.files.files.each {
+                        println("ERROR-PRONE: ${it.name}")
+                    }
+                }
+            }
+        '''.stripIndent(true)
+
+        when:
+        def output = runTasksSuccessfully('printErrorProneVersions').standardOutput
+
+        then: 'every single error-prone dependency has the same version'
+        output.contains('ERROR-PRONE: error_prone_annotation-2.3.4.jar')
+        output.contains('ERROR-PRONE: error_prone_core-2.3.4.jar')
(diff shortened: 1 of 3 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1226,6 +1226,34 @@ class SuppressibleErrorPronePluginIntegrationTest extends IntegrationSpec {
...
+            }
+        '''.stripIndent(true)
+
+        when:
+        def output = runTasksSuccessfully('printErrorProneVersions').standardOutput
+
+        then: 'every single error-prone dependency has the same version'
+        output.contains('ERROR-PRONE: error_prone_annotation-2.3.4.jar')
+        output.contains('ERROR-PRONE: error_prone_core-2.3.4.jar')
+    }
```
felixdesouza, line 1255: what would happen if you didn't have a force? is there a way to do a before and after style test? Or to recreate the scenario you're talking about

CRogers: I don't think I can, without adding a flag or something to similar to disable the ComponentMetadataRule that I've added. Now these versions will always be the same, I can't replicate the previous issue (as it's fixed).

felixdesouza: fair

felixdesouza:
👍

PR #114, opened 2025-07-16
Refactor to add the concept of `Mode`s and how they interfere with each other

CRogers:
## Before this PR
I was working on adding yet another mode to this project, something like `-PerrorProneRemoveUnusedRolloutSupressions` which will run at the same time as suppressing/applying suggesting fixes for efficiency.

However, the logic for setting patch options and other configuration in this repo is already *really* complicated and I struggle to wrap my head around it. We have a series of doubly nested ifs and complicated control flow to configure errorprone properly. In adding this new feature, I was about to make some of these triply nested ifs, and it all seemed a bit too much.

We really need to simplify the existing implementation to be less complicated before we add any more intrinsic complexity.

## After this PR
I've solidified the concepts of `Mode`s into classes. Each flag corresponds to a `Mode`, eg `-PerrorProneSupress` is handled by a `SuppressMode` that handles the configuration for this mode.

Most of the complexity in this repo comes from the fact that mode _interfere_ with each other. In particular, when you suppress and apply at the same time, you need to do something different to just suppressing or apply independently. This is where `ModeInterference` comes in - it can take the outputs of two `Mode`s and change their common configuration to work properly. This cuts down on the nested ifs entirely, and the main Gradle Plugin class is way simpler.

Modes are entirely independent and can be thought about independently unless they interfere, and that is handled by another class.

I'm releasing this as this is a large enough change I want to check it works across all our repos before merging in something built on top of it, so changelog:
==COMMIT_MSG==
Internal only change: Refactor to add the concept of `Mode`s and how they interfere with each other.
==COMMIT_MSG==

## Possible downsides?
While all the control flow and if nesting has gone, there are now more concepts and more files you need to look at the fully understand what is happening. I think this is better overall - the concepts were there before but implicit. The files allows you analyse different parts independently without having to worry about everything at once (they are hopefully a good abstraction).


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -191,190 +166,55 @@ public void execute(ComponentMetadataContext context) {
...
-        if (isRemovingSuppressions(project)) {
-            errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
-                @Override
-                public Iterable<String> asArguments() {
-                    List<String> suppressionsToRemove = checksToRemoveSuppressionsFor(project);
-
-                    Set<String> checksToPatch = new HashSet<>();
-                    checksToPatch.add("RemoveRolloutSuppressions");
-
-                    if (isApplyingSuggestedPatches(project)) {
```
CRogers, line 250: This `if` and below is an example of the complexity that's practically all gone.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
```diff
@@ -0,0 +1,59 @@
...
+        return ModifyCheckApiOption.dontCare();
+    }
+
+    /**
+     * Apply configuration specific to the mode or return the {@link ModeOptions} for this mode.
+     * @param context Useful information to use to determine options or apply configuration
+     */
+    default ModeOptions options(ModeOptionContext context) {
+        return DontCare.INSTANCE;
+    }
```
kelvinou01: `Mode::options` is both a query and mutation method — it 1. uses `context` as a source of information to determine `ModeOption`, while 2. simultaneously mutating `context`.  

To make it easier to reason with, how about we make `Mode::options` a pure function, while having another method `Mode::configure` dedicated to mutating the context?

```java
    /**
     * Return the {@link ModeOptions} for this mode.
     * @param readOnlyContext Useful information to use to determine options, shouldn't be mutated
     */
    default ModeOptions options(ModeOptionContext readOnlyContext) {
        return DontCare.INSTANCE;
    }

    /**
     * Apply configuration specific to the mode.
     * @param config Useful information to use to apply configuration, can be mutated.
     */
    default void configure(ModeOptionContext config) {
        // No configuration by default
    }
```


Then, `DisableMode` will look like

```java
public final class DisableMode implements Mode {

    @Override
    public void configure(ModeOptionContext config) {
        // Options (including historical backcompat from when this logic live in gradle-baseline):
        //   -PerrorProneDisable
        //   -Pcom.palantir.baseline-error-prone.disable
        //   -Pcom.palantir.baseline-error-prone.disable=true
        //   -Pcom.palantir.baseline-error-prone.disable=false
        // So if the value is true, we actually set errorProneOptions.enabled to false

        config.errorProneOptions()
                .getEnabled()
                .set(config.flagValue()
                        .map(value -> !Boolean.parseBoolean(value))
                        .orElse(false));
    }

    // use default implementation of `options`, which returns `DontCare.INSTANCE`
}
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,139 @@
...
+                    Mode mode = modes.get(flagName);
+                    return mode.options(new ModeOptionContext(flagValue, javaCompile));
+                })
+                .toMap();
+
+        Map<Set<ModeName>, ModeOptions> interferingModeOptions = EntryStream.of(interferingModes)
+                .mapToValue((interferingModeNames, modeInterference) -> {
+                    return modeInterference.interfere(StreamEx.of(interferingModeNames)
+                            .mapToEntry(modeOptions::get)
+                            .toMap());
```
kelvinou01: The nested stream operations in `modeOptionsFor()` are hard to follow. Consider extracting the inner logic:

```java
        Map<Set<ModeName>, ModeOptions> interferingModeOptions = EntryStream.of(interferingModes)
                .mapToValue((interferingModeNames, modeInterference) -> {
                    Map<ModeName, ModeOptions> nameToOption = StreamEx.of(interferingModeNames)
                            .mapToEntry(modeOptions::get)
                            .toMap();
                    return modeInterference.interfere(nameToOption);
                })
                .toMap();
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,139 @@
...
+                    if (flagValuesToNames.size() > 1) {
+                        throw new IllegalArgumentException(
+                                "Multiple instances of flags for the same mode were supplied with different values: "
+                                        + flagValuesToNames);
+                    }
+
+                    return Map.of(
+                            modeName, flagValuesToNames.keySet().iterator().next());
+                })
+                .mapValues(value -> Optional.of(value).filter(Predicate.not(String::isBlank)))
```
kelvinou01, line 155: We can use guava's [`Strings.emptyToNull(value)`](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html#emptyToNull-java.lang.String-) (unless if we also want to filter out whitespace, which im not sure we want to?)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
```diff
@@ -0,0 +1,44 @@
...
+    Set<ModeName> interferesWith(Set<ModeName> modeNames);
+
+    /**
+     * Modify the modeOptions if the flags interfere with each other.
+     * @param modeOptions A map from the {@link ModeName}s that were indicated to be interfering in
+     *                    {@link #interferesWith} to their corresponding {@link ModeOptions}.
+     * @return The modified {@link ModeOptions} that will be used in place of the originals that the flags produced
+     */
+    default ModeOptions interfere(Map<ModeName, ModeOptions> modeOptions) {
+        throw new UnsupportedOperationException("interfere is not implemented");
```
kelvinou01: Maybe `IllegalStateException("this interference is not permitted")` conveys your intent better?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
```diff
@@ -0,0 +1,44 @@
...
+ * Each of these provide a way to indicate which {@link ModeName}s interfere with each other and modify
+ * the {@link ModeOptions} they produce in order to work properly together.
+ */
+public interface ModeInterference {
+    /**
+     * Identify which of the flags interfere with each other.
+     * @param modeNames The flags that are enabled in this run
+     * @return Either the flags that interfere with each other, or throw an exception if they are incompatible
+     */
+    Set<ModeName> interferesWith(Set<ModeName> modeNames);
```
kelvinou01: If it will only either be zero or a pair of modes that interfere with one another, consider narrowing the return type to `Optional<Pair<ModeName, ModeName>>`?

This will make it even easier to reason with `ModeInterference`, and simplify the logic in `Modes::modeOptionsFor`.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
```diff
@@ -0,0 +1,53 @@
...
+
+            @Override
+            public Map<String, String> extraErrorProneCheckFlags() {
+                // For the suppressions to remove, if no specific check is enabled, we need to just remove everything
+                // We can't explicitly list all possible checks, because some might not exist anymore
+                // The logic itself needs to consider an empty list as "remove all"
+
+                return Map.of(
+                        "SuppressibleErrorProne:RemoveRolloutSuppressions",
+                        context.flagValue().orElse(""));
```
kelvinou01: There's multiple instances here where we use `""` to mean "all checks". The comments explain it well, but would it be worth it to define a constant, maybe `CHECK_FLAG_ALL_CHECKS = ""`?

kelvinou01 (commented):
Very cool PR! It's a big reduction in complexity. 

I'm new to this project and I don't know what the new mode you're working on will entail, so let me know if any of my assumptions are wrong.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
CRogers: Nice idea! I actually did this to begin with, I can't remember why I moved away from it now... Will try to make it like this again.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
CRogers: Yeah I think making a "real type" using sealed interfaces like elsewhere might make sense here.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
CRogers: I tried it but it makes the implementation more error prone... you need to both remember to get the common options and call `configure` on each of them. I've updated the naming to hopefully make it more obvious - lmk what you think.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
CRogers: Done

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
CRogers: I think I wrote this just so that I didn't have to add empty/throwing implementations in `DisableModeInterference`. It's a bit of special case so I'll just implement this there and make it required

CRogers: Actually I think it does make sense - I've made the error message clearer what's going on here.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
```diff
@@ -0,0 +1,112 @@
...
+
+        @Override
+        public boolean anyChecks() {
+            return true;
+        }
+
+        @Override
+        public Optional<String> asCommaSeparated() {
+            // Empty string means "all checks" to Error Prone - ie the option `-XepPatchChecks:`
+            return Optional.of("");
```
kelvinou01, line 79: Let's do `ALL_CHECKS` here?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/SuppressingAndApplyingInterference.java
```diff
@@ -0,0 +1,48 @@
...
+        // errorprone to patch all checks, so we can make suggested fixes for suppressions in any check.
+        // However, inside our changes to errorprone, we need to get the list of checks that we're going
+        // to use the default suggested fixes for, so we can work out which ones to use the suggested
+        // fixes for and which to suppress. So we add the PreferPatchChecks argument here, which we can
+        // use inside error-prone/the compiler.
+
+        return suppress.naivelyCombinedWith(apply)
+                .withExtraErrorProneCheckFlag(
+                        "SuppressibleErrorProne:PreferPatchChecks",
+                        apply.patchChecks().asCommaSeparated().orElse(""));
```
kelvinou01, line 46: `ALL_CHECKS` here as well?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
```diff
@@ -0,0 +1,55 @@
...
+package com.palantir.gradle.suppressibleerrorprone.modes.modes;
+
+import com.palantir.gradle.suppressibleerrorprone.modes.common.CommonOptions;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.Mode;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.ModifyCheckApiOption;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.PatchChecksOption;
+import java.util.Map;
+
+public final class RemoveRolloutMode implements Mode {
+    private static final String ALL_CHECKS = "";
```
kelvinou01, line 26: Let's explain here that errorprone considers "" as everything?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        enum NoInterference implements ModeInterferenceResult {
+            INSTANCE
+        }
+
+        record Interference(Set<ModeName> interferingModes) implements ModeInterferenceResult {
+            public Interference {
+                Preconditions.checkArgument(
+                        interferingModes.size() > 1, "interference must be between at least 2 modes");
```
kelvinou01: Can there be interference with >2, e.g. 3 modes?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
kelvinou01: > you need to both remember to get the common options and call configure on each of them.

do you mean that having to do both `options(context)` and `configure(context)` is error-prone?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
CRogers: yes

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -0,0 +1,95 @@
...
+            @Override
+            public Map<String, String> extraErrorProneCheckFlags() {
+                Map<String, String> map = new HashMap<>(super.extraErrorProneCheckFlags());
+                map.put(key, value);
+                return map;
+            }
+        };
+    }
+
+    static CommonOptions dontCare() {
```
felixdesouza: I think you should refer to this as `.empty`. My initial read was that `dontCare` would allow duplicate keys in a sense. However, there's no handling of duplicate keys here, it will blow up if so.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -0,0 +1,95 @@
...
+            @Override
+            public PatchChecksOption patchChecks() {
+                return CommonOptions.this.patchChecks().combine(other.patchChecks());
+            }
+
+            @Override
+            public Map<String, String> extraErrorProneCheckFlags() {
+                return EntryStream.of(CommonOptions.this.extraErrorProneCheckFlags())
+                        .append(other.extraErrorProneCheckFlags())
+                        .toMap();
```
felixdesouza, line 50: Do you care about duplicate key elements? It's going to explode here if so.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -0,0 +1,95 @@
...
+    static CommonOptions dontCare() {
+        return DontCare.INSTANCE;
+    }
+
+    enum DontCare implements CommonOptions {
+        INSTANCE
+    }
+
+    @SuppressWarnings("checkstyle:DesignForExtension")
+    class DefaultCommonOptions implements CommonOptions {
```
felixdesouza: surprised you didn't use a `record` for this?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,92 @@
...
+    static ModifyCheckApiOption mustModifyIncludingVisitorState() {
+        return new MustModify(true);
+    }
+
+    sealed interface CombinedValue permits DoNotModify, MustModify {}
+
+    static CombinedValue combine(Collection<ModifyCheckApiOption> options) {
+        Set<ModifyCheckApiOption> withoutDontCares = options.stream()
+                .filter(Predicate.not(Predicate.isEqual(DontCare.INSTANCE)))
+                .collect(Collectors.toSet());
```
felixdesouza: Think you can just do:

```java
StreamEx.of(options)
  .remove(DontCare.INSTANCE)
  .toSet();
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,92 @@
...
+
+        if (withoutDontCares.isEmpty()) {
+            // By default, we need to modify the check API to support for-rollout suppressions
+            return new MustModify(false);
+        }
+
+        boolean doNotModify = withoutDontCares.contains(DoNotModify.INSTANCE);
```
felixdesouza: I haven't fully reached yet, so this question is probably going to be ill-defined, but isn't this basically an interference?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * suppressible-error-prone is powered by an artifact transform that modifies the error_prone_check_api jar.
+ * Some modes must not run with this jar modified, some must run with it modified (and then some may or may not need
+ * the VisitorState class modified), and some don't care. This class represents these states and allows combining them
+ * sensibly.
+ */
+public sealed interface ModifyCheckApiOption permits DoNotModify, DontCare, MustModify {
```
felixdesouza: I like the sealed interface pattern, but it does make the class rather noisy. Can we move them to be separate files, or put them in a single private class/interface file. Claude seems to suggest that the classes referenced in the permits can be private. Then we can just focus on the static constructors you have here and expect people to use.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+        };
+    }
+
+    static CommonOptions naivelyCombine(Collection<CommonOptions> commonOptions) {
+        return commonOptions.stream().reduce(CommonOptions.dontCare(), CommonOptions::naivelyCombinedWith);
+    }
+
+    default CommonOptions withExtraErrorProneCheckFlag(String key, String value) {
+        return new DefaultCommonOptions(this) {
```
felixdesouza: why not do the map construction and then pass it into the constructor? then you don't need a normal class for `DefaultCommonOptions`, and can just use a record.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
```diff
@@ -0,0 +1,112 @@
...
+
+    default PatchChecksOption combine(PatchChecksOption other) {
+        if (this instanceof AllChecks || other instanceof AllChecks) {
+            return AllChecks.INSTANCE;
+        }
+
+        if (this instanceof PossiblySomeChecks thisSome && other instanceof PossiblySomeChecks otherSome) {
+            return new PossiblySomeChecks(
+                    () -> Stream.concat(thisSome.patchChecks().stream(), otherSome.patchChecks().stream())
+                            .collect(Collectors.toSet()));
```
felixdesouza: nit: `StreamEx.of(thisSome.patchChecks()).append(otherSome.patchChecks()).toSet()`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        throw new IllegalArgumentException("Must be an instance of AllChecks or SomeChecks");
```
felixdesouza, line 49: technically unreachable yes

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    final class PossiblySomeChecks implements PatchChecksOption {
+        private final Supplier<Set<String>> patchChecks;
+
+        public PossiblySomeChecks(Supplier<Set<String>> patchChecks) {
+            // Memoize as if this is called at different times in the build, different values could be returned.
+            // At least try to keep it consistent.
+            this.patchChecks = Suppliers.memoize(() -> Set.copyOf(patchChecks.get()));
```
felixdesouza, line 89: why not just not have the supplier and just take the first option you get? I have not reached the part where this is called just yet, so we'll see.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        @Override
+        public Optional<String> asCommaSeparated() {
+            // Empty string means "all checks" to Error Prone - ie the option `-XepPatchChecks:`
+            return Optional.of("");
+        }
+    }
+
+    final class PossiblySomeChecks implements PatchChecksOption {
```
felixdesouza, line 83: similarly, let's spin this out into a separate class, or a holder interface. Same for `AllChecks`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
felixdesouza: I think, it's probably actually clearer if you implement the `combine` method on the actual implementations:

```java
AllChecks {
  PatchChecksOption combine(PatchChecksOption other) {
    return AllChecks.INSTANCE;
  }
}

PossiblySomeChecks {
  PatchCheckOption combine(PatchChecksOption other) {
    if (other instanceof AllChecks allChecks) {
      return allChecks.combine(this);
    } else if (other instanceof PossiblySomeCheck possiblySomeChecks) {
      return StreamEx.of(thisSome.patchChecks()).append(possiblySomeChecks.patchChecks()).toSet()
    }
  }
}
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/DisableModeInterference.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public ModeInterferenceResult interferesWith(Set<ModeName> modeNames) {
+        if (modeNames.contains(ModeName.DISABLE) && modeNames.size() > 1) {
+            throw new IllegalStateException("%s cannot be used at the same time as any of %s"
+                    .formatted(
+                            ModeName.DISABLE.asGradlePropertyArgument(),
+                            modeNames.stream()
+                                    .filter(Predicate.not(Predicate.isEqual(ModeName.DISABLE)))
+                                    .map(ModeName::asGradlePropertyArgument)
+                                    .collect(Collectors.joining(", "))));
```
felixdesouza: nit:
```java
StreamEx.of(modeNames).remove(ModeName.DISABLE).map(ModeName::asGradlePropertyArgument).joining(", ")
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemovingAndSuppressingInterference.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.suppressibleerrorprone.modes.interferences;
+
+import com.palantir.gradle.suppressibleerrorprone.modes.common.ModeInterference;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.ModeName;
+import java.util.Set;
+
+public final class RemovingAndSuppressingInterference implements ModeInterference {
```
felixdesouza, line 28: A thought, it'd be helpful to explain why this mode interference is the way it is? Maybe not necessarily in the error message, but as a docstring perhaps, idk. Largely from the perspective of a dev who may want to add new modes, and needs to evaluate whether or not the new mode interferes or not. If the reason for the interference changes, then this can be re-evaluated.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemovingAndSuppressingInterference.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public ModeInterferenceResult interferesWith(Set<ModeName> modeNames) {
+        if (modeNames.containsAll(Set.of(ModeName.REMOVE_ROLLOUT, ModeName.SUPPRESS))) {
+            throw new IllegalStateException("%s cannot be used at the same time as %s"
```
felixdesouza: being more pedantic more than anything, but it's a little weird that the fact that two modes are incompatible do not get featured in the result? I would have thought it'd be a separate state.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
felixdesouza: we should move this is somewhere else and reference it (it's used in a couple of places) or just copy paste the field and explanation in those places.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public ModifyCheckApiOption modifyCheckApi() {
+        // If we're going to remove suppressions, and possibly apply patches, we don't want to apply the custom
+        // logic for for-rollout suppressions.
```
felixdesouza, line 31: I think we should flesh this out a bit more, we don't talk about what `ModifyCheckApiOption` might actually do in its docs, just that some modes need it/explicitly don't want it. This mode needs it to not be modified, but the why isn't super obvious from this comment. I'm still not sure I understand it myself 😅

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/SuppressMode.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public CommonOptions configureAndReturnCommonOptions(ModeOptionContext context) {
+        return new CommonOptions() {
+            @Override
+            public PatchChecksOption patchChecks() {
+                return PatchChecksOption.allChecks();
+            }
+        };
+    }
```
felixdesouza, line 38: separately, why go for the anonymous class approach rather than just a builder?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+                .collect(Collectors.toSet()));
+    }
+
+    public final CommonOptions commonOptionsFor(JavaCompile javaCompile) {
+        Map<ModeName, Optional<String>> modeNameToFlagValue = modesEnabledAndFlagValues();
+
+        Map<Interference, ModeInterference> interferingModes = StreamEx.of(interferences)
+                .mapToEntry(interference -> interference.interferesWith(modeNameToFlagValue.keySet()))
+                .filterValues(modeInterferenceResult -> modeInterferenceResult instanceof Interference)
+                .mapValues(modeInterferenceResult -> (Interference) modeInterferenceResult)
```
felixdesouza: you can replace this with:

```suggestion
                .selectValues(Interference.class)
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+                .mapToValue((interference, modeInterference) -> {
+                    return modeInterference.interfere(StreamEx.of(interference.interferingModes())
+                            .mapToEntry(commonOptionsPerMode::get)
+                            .toMap());
+                })
+                .toMap();
+
+        Set<ModeName> allInterferingModes = interferingModes.keySet().stream()
+                .flatMap(interference -> interference.interferingModes().stream())
+                .collect(Collectors.toSet());
```
felixdesouza: nit: can we just use `StreamEx` throughout?


```suggestion
        Set<ModeName> allInterferingModes = StreamEx.of(interferingModes.keySet())
                .flatMap(interference -> interference.interferingModes().stream())
                .toSet();
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+
+        Set<CommonOptions> nonInterferingCommonOptions = EntryStream.of(commonOptionsPerMode)
+                .filterKeys(Predicate.not(allInterferingModes::contains))
```
felixdesouza:
```suggestion
                .removeKeys(allInterferingModes::contains)
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+                .values()
+                .toSet();
+
+        Set<CommonOptions> allCommonOptions = StreamEx.of(commonOptionsAfterInterfering.values())
+                .append(nonInterferingCommonOptions)
+                .collect(Collectors.toSet());
```
felixdesouza:
```suggestion
                .toSet();
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+                    }
+
+                    if (flagValuesToNames.size() > 1) {
+                        throw new IllegalArgumentException(
+                                "Multiple instances of flags for the same mode were supplied with different values: "
+                                        + flagValuesToNames);
+                    }
+
+                    return Map.of(
+                            modeName, flagValuesToNames.keySet().iterator().next());
```
felixdesouza: can we ensure this is consistent ordering? perhaps it matters less in different execution environments, but it would suck if wrecked build cache because the order is not consistent

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
felixdesouza: > We can use guava's [`Strings.emptyToNull(value)`](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html#emptyToNull-java.lang.String-) (unless if we also want to filter out whitespace, which im not sure we want to?)

Depending on intent, this is not the same thing?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
```diff
@@ -0,0 +1,142 @@
...
+            ModeName.SUPPRESS, new SuppressMode(),
+            ModeName.TIMINGS, new TimingsMode());
+
+    private final Set<ModeInterference> interferences = Set.of(
+            new DisableModeInterference(),
+            new RemovingAndSuppressingInterference(),
+            new SuppressingAndApplyingInterference());
+
+    public final CombinedValue modifyCheckApi() {
+        return ModifyCheckApiOption.combine(modesEnabledAndFlagValues().keySet().stream()
```
felixdesouza: nit: streamex for this pls

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Apply configuration specific to the mode or return the {@link CommonOptions} for this mode.
+     * @param context Useful information to use to determine options or apply configuration
+     */
+    default CommonOptions configureAndReturnCommonOptions(ModeOptionContext context) {
+        return DontCare.INSTANCE;
+    }
+
+    record ModeOptionContext(Optional<String> flagValue, JavaCompile javaCompile) {
```
felixdesouza, line 46: can we explain what the empty flag value is supposed to represent?

without reading the source, it's not immediateyl clear whether this is called for *all* modes, and you filter based on `flagValue`, or for the actual impl, this is called by filtering the modes supplied, and then this is a potential value or not.

I would not be opposed to having another helper interface if that's also easier

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -191,190 +166,55 @@ public void execute(ComponentMetadataContext context) {
...
-            javaCompile.getOutputs().cacheIf(t -> false);
-            javaCompile.getOutputs().upToDateWhen(t -> false);
+    private static void configureJavaCompile(CommonOptions commonOptions, JavaCompile javaCompile) {
+        // Don't attempt to cache or be up-to-date since it won't capture the source files that might be modified
+        javaCompile.getOutputs().cacheIf(t -> !commonOptions.patchChecks().anyChecks());
+        javaCompile.getOutputs().upToDateWhen(t -> !commonOptions.patchChecks().anyChecks());
+    }
+
+    private static void afterEvaluateConfigureJavaCompile(CommonOptions commonOptions, JavaCompile javaCompile) {
+        if (commonOptions.patchChecks().anyChecks()) {
```
felixdesouza: nit: reading this and also the `PatchChecksOptions`, I feel this reads better as `hasChecks()`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -191,190 +166,55 @@ public void execute(ComponentMetadataContext context) {
...
-
-        // For the suppressions to remove, if no specific check is enabled, we need to just remove everything
-        // We can't explicitly list all possible checks, because some might not exist anymore
-        // The logic itself needs to consider an empty list as "remove all"
-        if (possibleChecksToRemove == null) {
-            return List.of();
-        }
+        errorProneOptions
+                .getCheckOptions()
+                .putAll(getProviderFactory().provider(commonOptions::extraErrorProneCheckFlags));
```
felixdesouza: can we call this the same thing: `extraErrorProneCheckOptions`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -191,190 +166,55 @@ public void execute(ComponentMetadataContext context) {
...
-                        List<String> patchChecks =
-                                checksToApplySuggestedPatchesFor(extension, javaCompile, errorProneOptions);
-
-                        return List.of(
-                                "-XepOpt:SuppressibleErrorProne:PreferPatchChecks=" + String.join(",", patchChecks));
-                    }
-                });
+        errorProneOptions.getErrorproneArgumentProviders().add(new CommandLineArgumentProvider() {
+            @Override
+            public Iterable<String> asArguments() {
```
felixdesouza, line 194: if we're already in an `afterEvaluate` do we actually need this to be a `argumentProvider` ? Can't we just pass this in directly to `getErrorproneArgs`?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
felixdesouza: same question, does this actually need to be a provider since we're not in an `afterEvaluate`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -191,190 +166,55 @@ public void execute(ComponentMetadataContext context) {
...
-            });
-        }
-
-        if (isAnyKindOfPatching(project)) {
-            // Don't attempt to cache or be up-to-date since it won't capture the source files that might be modified
-            javaCompile.getOutputs().cacheIf(t -> false);
-            javaCompile.getOutputs().upToDateWhen(t -> false);
+    private static void configureJavaCompile(CommonOptions commonOptions, JavaCompile javaCompile) {
+        // Don't attempt to cache or be up-to-date since it won't capture the source files that might be modified
+        javaCompile.getOutputs().cacheIf(t -> !commonOptions.patchChecks().anyChecks());
```
felixdesouza: I wonder whether this is CC safe

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/Mode.java
```diff
@@ -0,0 +1,59 @@
...
+    default ModifyCheckApiOption modifyCheckApi() {
+        return ModifyCheckApiOption.dontCare();
+    }
+
+    /**
+     * Apply configuration specific to the mode or return the {@link CommonOptions} for this mode.
+     * @param context Useful information to use to determine options or apply configuration
+     */
+    default CommonOptions configureAndReturnCommonOptions(ModeOptionContext context) {
+        return DontCare.INSTANCE;
```
felixdesouza: I feel like we have a `DontCare` in different contexts, and it's a little confusing, can we have a better name for those cases pls 🙏

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
CRogers: Maybe I'm missing something - but what would streamex provide here? I'm just doing basic maps no?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
CRogers: We aren't using any streamex specific methods though? Is this just for better consistency? I kinda like avoiding using it unless there's a good reason to, streamex have had a number of abi breaks before and the more usages the more we risk hitting them.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
CRogers: nice

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
CRogers: There is only one element in this map so it does not matter. I've changed it use getOnlyElement to make it obvious (I think I had not introduced guava before I wrote this)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
CRogers: I think this desired behaviour. If multiple modes produce the same flag but with different values, something needs to resolve that. Are you suggesting we should have specific error handling for this though?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
CRogers: I think it makes sense all in one class, as most of the stuff inside is implementation details - I'm not sure putting them in another class really adds much except spreading it across two classes?. Perhaps I can just move the static constructors to the top of the class?

CRogers: I think of these like Algebraic Data Types in a language like haskell or rust, where you list everything together:

```haskell
data ModifyCheckOptionApi
    = DoNotModify
    | DontCare
    | MustModify Boolean
```

Spreading them out across multiple files feels a bit much to me? I would kind of understand if the intent was for people to add their own implementations - but they literally can't.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemovingAndSuppressingInterference.java
CRogers: I think have a separate state you have to deal with and but then basically throw immediately just makes `Modes` more complicated, and it's already complicated enough.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
CRogers: Because then this will calculate the patch checks too early before any other plugins have had a chance to configure it.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: I have made `CommonOptions` serializable so it _should_ be.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/PatchChecksOption.java
CRogers: As discussed in person, this makes the interface weird - you have to refer to `PatchChecksOptionImpl.PossiblySomeChecks` rather than `PatchChecksOption.PossiblySomeChecks` - would rather not reveal in the impl in the types on the interface.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
CRogers: This is actually entirely separate to the errorprone `-Xep:PatchChecks` parameter. This is some internal communication between our gradle plugin and the code running inside the compiler just for the `RemoveRollout` mode. They both just happen to define empty string to mean all the checks. They aren't necessarily related.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemovingAndSuppressingInterference.java
CRogers: I changed it so being not compatible is part of the result now rather than randomly throwing

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/SuppressMode.java
CRogers: As discussed, it's helpful in debugging to see where the implementations come from. Also means don't need to apply immutables.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/Modes.java
CRogers: I have used streamex throughout.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
CRogers: Might as well make it a provider rather than calling it immediately?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModeInterference.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Identify which of the flags interfere with each other.
+     *
+     * @param modeNames The flags that are enabled in this run
+     * @return Either no interference, the flags that interfere with each other, or why they are incompatible
+     */
+    ModeInterferenceResult interferesWith(Set<ModeName> modeNames);
+
+    /**
+     * Modify the modeCommonOptions if the flags interfere with each other.
```
felixdesouza, line 37:
```suggestion
     * Modify the {@code modeCommonOptions} if the flags interfere with each other.
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * suppressible-error-prone is powered by an artifact transform that modifies the error_prone_check_api jar.
+ * Some modes must not run with this jar modified, some must run with it modified (and then some may or may not need
+ * the VisitorState class modified), and some don't care. This class represents these states and allows combining them
+ * sensibly.
+ */
+public sealed interface ModifyCheckApiOption permits DoNotModify, NoEffect, MustModify {
+    /**
+     * The error_prone_check_api jar must not be modified.
```
felixdesouza:
```suggestion
     * The {@code error_prone_check_api} jar must not be modified.
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    static ModifyCheckApiOption doNotModify() {
+        return DoNotModify.INSTANCE;
+    }
+
+    /**
+     * It doesn't matter whether the error_prone_check_api jar is modified or not. This will have no effect on the
```
felixdesouza:
```suggestion
     * It doesn't matter whether the {@code error_prone_check_api} jar is modified or not. This will have no effect on the
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,106 @@
...
+     * decision made.
+     */
+    static ModifyCheckApiOption noEffect() {
+        return NoEffect.INSTANCE;
+    }
+
+    /**
+     * The error_prone_check_api jar must be modified to allow `for-rollout:` suppressions to work.
```
felixdesouza:
```suggestion
     * The {@code error_prone_check_api} jar must be modified to allow {@code for-rollout:} suppressions to work.
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    static ModifyCheckApiOption mustModify() {
+        return new MustModify(false);
+    }
+
+    /**
+     * The error_prone_check_api jar must be modified to allow `for-rollout:` suppressions to work, and VisitorState
```
felixdesouza:
```suggestion
     * The{@code error_prone_check_api} jar must be modified to allow {@code for-rollout:} suppressions to work, and {@code VisitorState}
```

I'm not sure if `VisitorState` is on the classpath, if so `{@link VisitorState}` instead?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
felixdesouza, line 31:
```suggestion
        // logic for {@code for-rollout} suppressions.
```

CRogers: This isn't javadoc?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/TimingsMode.java
```diff
@@ -0,0 +1,55 @@
...
+                .get()
+                .getAsFile()
+                .toPath();
+
+        context.javaCompile().getOutputs().file(outputAbsolute.toFile());
+
+        context.javaCompile().getOptions().getCompilerArgumentProviders().add(new CommandLineArgumentProvider() {
+            @Override
+            public Iterable<String> asArguments() {
+                return List.of("-Xplugin:SuppressibleErrorProneTimings " + outputAbsolute);
```
felixdesouza, line 49: shouldn't this be two elements? also you can just apply this as args no? there's no laziness here

CRogers: I think we should be using `getCompilerArgumentProviders()` rather than `getCompilerArgs`/`setCompilerArgs` - the latter is old api

felixdesouza: oh, I did not realise that it wasn't a property, that's pretty grim

CRogers: This is correct as is, it's one argument. List with two elements:

```
Caused by: java.lang.IllegalArgumentException: error: invalid flag: /Users/callumr/git/suppressible-error-prone/gradle-suppressible-error-prone/build/nebulatest/com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePluginIntegrationTest/timings-are-outputted/build/errorprone-timings/compileJava
	at jdk.compiler/com.sun.tools.javac.main.Arguments.reportDiag(Arguments.java:886)
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
felixdesouza: oop

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/TimingsMode.java
felixdesouza: oh this is calling javac through the class and not via exec, alright

felixdesouza:
👍

PR #136, opened 2025-08-21
Configuration Cache compatibility

CRogers:
## Before this PR
This plugin is not compatible with configuration cache after #114. You get this lovely error:

> Could not load the value of field `interferences` of `com.palantir.gradle.suppressibleerrorprone.modes.Modes_Decorated` bean found in field `__modes__` of `com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin_Decorated` bean found in field `this$0` of `com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin$1` bean found in field `errorproneArgumentProviders` of `net.ltgt.gradle.errorprone.ErrorProneOptions_Decorated` bean found in field `errorproneOptions` of `net.ltgt.gradle.errorprone.ErrorProneCompilerArgumentProvider` bean found in field `compilerArgumentProviders` of `org.gradle.api.tasks.compile.CompileOptions_Decorated` bean found in field `capturedArgs` of `java.lang.invoke.SerializedLambda` bean found in field `spec` of `org.gradle.api.internal.tasks.execution.SelfDescribingSpec` bean found in task `:compileJava` of type `org.gradle.api.tasks.compile.JavaCompile`

The problem is we inject an object into the `CommandLineArgumentProvider` that is not fully `Serializable`. This object is more information than is necessary.

Additionally, there is another issue where `getProject()` is called at task execution time to calculate the `cacheIf` and `upToDateWhen` specs:

```
- Task `:compileJava` of type `org.gradle.api.tasks.compile.JavaCompile`: invocation of 'Task.project' at execution time is unsupported.
- Task `:compileOtherJava` of type `org.gradle.api.tasks.compile.JavaCompile`: invocation of 'Task.project' at execution time is unsupported.
- Task `:compileTestJava` of type `org.gradle.api.tasks.compile.JavaCompile`: invocation of 'Task.project' at execution time is unsupported.
```

## After this PR
==COMMIT_MSG==
This plugin now supports Configuration Cache.
==COMMIT_MSG==

To fix, we make a managed `CommandLineArgumentProvider` that we can lazily set the value we need, and Gradle knows how to handle this properly for configuration cache.

To fix the `getProjects`, we stop trying to detect if there are actually 0 checks being patched (which involves doing some filtering logic that depends on calling `getProject`) and instead just know whether we're in a patching mode or not.


suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/FirstTimeMemoizingFunction.java
```diff
@@ -18,14 +18,12 @@
...
 import java.util.Optional;
 import java.util.function.Function;
-import javax.annotation.concurrent.NotThreadSafe;
 
 /**
  * Memoizes the first evaluation of a function with one parameter. Subsequent evaluations will return the value
  * calculated in the first run, even if the argument used in subsequent evaluations differs from the argument used
  * in the first evaluation.
  */
-@NotThreadSafe
```
felixdesouza, line 28: why is this no longer true?

FinlayRJW, line 28: why is this now thread safe?

CRogers: It isn't - guava updated and removed the jsr-whatever dep that had this, so I just removed it - it's just advisory anyway, and the places in Gradle this is used is single threaded.

felixdesouza (approved):

PR #141, opened 2025-08-28
Ensure running `-PerrorProneSuppress` and `-PerrorProneApply` at the same time do not resolve Configurations at configuration time

CRogers:
## Before this PR
We noticed that our excavators stopped applying errorprone fixes. When running with both `-PerrorProneSuppress` and `-PerrorProneApply` at the same time, and a `IfModuleIsUsed` conditional patch check is registered (such as the one baseline uses), we end up resolving the compile classpath Configuration for each source set at configuration time, which GCV bans:

```
org.gradle.api.GradleException: Not allowed to resolve configuration ':project:integrationTestCompileClasspath' at configuration time (https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time). Please upgrade your plugins and double-check your gradle scripts (see stacktrace)
```

Our tests didn't catch this because:

1. We did not apply GCV in the tests
2. No test that has `-PerrorProneApply`, `-PerrorProneSuppress` **and** a `IfModuleIsUsed` conditional patch check.

The problem was in the `SuppressingAndApplyingInterference`: we calculate the value of the `patchChecks()` for the apply mode immediately rather than lazily, at configuration time. When an `IfModuleIsUsed` conditional patch check is in use, it will resolve the compile classpath Configuration to see whether the module is used when calculating `patchChecks()`, at configuration time, causing the GCV error.

## After this PR
==COMMIT_MSG==
Ensure running `-PerrorProneSuppress` and `-PerrorProneApply` at the same time when an `IfModuleIsUsed` conditional patch check is in use do not resolve Configurations at configuration time, which cause this error when used with `gradle-consistent-versions`:

```
org.gradle.api.GradleException: Not allowed to resolve configuration ':project:integrationTestCompileClasspath' at configuration time (https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time). Please upgrade your plugins and double-check your gradle scripts (see stacktrace)
```

==COMMIT_MSG==

Also verified ETE on an internal project.

### DIFF
--- a/gradle-suppressible-error-prone/build.gradle
+++ b/gradle-suppressible-error-prone/build.gradle
@@ -52,3 +52,8 @@ tasks.withType(JavaCompile) {
 gradleTestUtils {
     gradleVersions = ['8.14.3']
 }
+
+moduleJvmArgs {
+    // Need so that debugging testkit gradle runs with configuration cache works
+    opens 'java.base/java.util', 'java.base/java.lang.invoke'
+}
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/SuppressingAndApplyingInterference.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/SuppressingAndApplyingInterference.java
@@ -43,6 +43,6 @@ public final class SuppressingAndApplyingInterference extends SpecificModeInterf
         return suppress.naivelyCombinedWith(apply)
                 .withExtraErrorProneCheckFlag(
                         "SuppressibleErrorProne:PreferPatchChecks",
-                        apply.patchChecks().asCommaSeparated().orElse(""));
+                        () -> apply.patchChecks().asCommaSeparated().orElse(""));
     }
 }
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -41,6 +41,18 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
 
         // language=Gradle
         buildFile << '''
+            buildscript {
+                repositories {
+                    mavenCentral()
+                }
+                dependencies {
+                    classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:3.1.0'
+                }
+            }
+            // Consistent versions checks we don't resolve configurations at configuration time and
+            // also interacts in many ways with dependencies
+            apply plugin: 'com.palantir.consistent-versions'
+
             apply plugin: 'com.palantir.suppressible-error-prone'
             apply plugin: 'java'
             
@@ -102,6 +114,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
             __TESTING=true
             __TESTING_CACHE_BUST_ERRORPRONE_TRANSFORM=true
         '''.stripIndent(true)
+
+        file('versions.lock')
     }
 
     def 'reports a failing error prone'() {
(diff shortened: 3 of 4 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/SuppressingAndApplyingInterference.java
```diff
@@ -43,6 +43,6 @@ public CommonOptions interfere(Map<ModeName, CommonOptions> modeCommonOptions) {
         return suppress.naivelyCombinedWith(apply)
                 .withExtraErrorProneCheckFlag(
                         "SuppressibleErrorProne:PreferPatchChecks",
-                        apply.patchChecks().asCommaSeparated().orElse(""));
```
CRogers, line 46: This was not evaluated lazily before.

felixdesouza: does the value of flag need to be evaluated lazily? I thought it would be sufficient with the wrapping you did above of the alternate method

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -41,6 +41,18 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
 
         // language=Gradle
         buildFile << '''
+            buildscript {
+                repositories {
+                    mavenCentral()
+                }
+                dependencies {
+                    classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:3.1.0'
```
felixdesouza, line 49: yeah, I am struggling to see a reason to not have gcv on stuff generally, I too have written a number of tests that just would not work where it's deployed.

CRogers: Yeah, every plugin should have GCV, it interacts with everything. Perhaps we should make sure any upcoming testing framework we have includes it by default.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -102,6 +114,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
             __TESTING=true
             __TESTING_CACHE_BUST_ERRORPRONE_TRANSFORM=true
         '''.stripIndent(true)
+
+        file('versions.lock')
```
felixdesouza, line 118: do we not need to `--write-locks` at any point?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/SuppressingAndApplyingInterference.java
CRogers: Calling `patchChecks()` is the issue here - so that needs be evaluated lazily. The value of flag is calculated from that so also needs to be lazy.

felixdesouza: sorry, read the PR description more closely, initially assumed the issue was in `extraErrorProneCheckOptions`

gradle-suppressible-error-prone/build.gradle
```diff
@@ -52,3 +52,8 @@ tasks.withType(JavaCompile) {
 gradleTestUtils {
     gradleVersions = ['8.14.3']
 }
+
+moduleJvmArgs {
```
felixdesouza, line 56: separate FLUP, but we should just add this to `gradle-test-utils` or whatever the plugin we have is called

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
felixdesouza: yep

felixdesouza:
👍

PR #143, opened 2025-08-29
Fix `RemoveRolloutSuppressions` not being applied on target repositories with `com.google.errorprone>=2.37.0`

kelvinou01:
## Before this PR

`RemoveRolloutSuppressions` stopped working for a lot of our repositories. It became a no-op.

We have a [test](https://github.com/palantir/suppressible-error-prone/blob/663fb789c10ec92c5a85abbc27e3cd9f83014434/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy#L982) which was supposed to catch this. But a double negative caused the test to pass anyways. 

### Root cause

The first mistake is when we introduced `Modes` in #114, we always set severity of `RemoveRolloutSuppressions` to `OFF`. This is wrong, because we wanna enable it when removing rollout suppressions. 

The upshot is that in `gw compileAllErrorProne -PerrorProneRemoveRollout`, these flags are supplied to errorprone:

1. `-Xep:RemoveRolloutSuppressions:OFF` 
2. `-XepPatchChecks:RemoveRolloutSuppressions`

The second mistake is in errorprone < 2.37.0, setting the severity of a patch check to `OFF` didn't actually turn it off. This was only fixed in [2.37.0](https://github.com/google/error-prone/releases/tag/v2.37.0)

> So patching now ignores checks:
> - Whose severity is OFF, or
> - That are not explicitly enabled for patching, while other checks are.

suppressible-error-prone is on errorprone 2.31.0, meaning that `RemoveRolloutSuppressions` ran despite it's severity being set to `OFF` erroneously, allowing the test to slyly pass. 

However, some of our other repositories are on e.g. errorprone 2.40.0. On those repositories, `RemoveRolloutSuppressions` didn't run (which is expected).

## After this PR

- Bump errorprone to 2.37.0 to bring in the correct interaction between patch checks and severity
- Only set `RemoveRolloutSuppressions` to `OFF` when we aren't removing rollout suppressions

==COMMIT_MSG==
Fix `RemoveRolloutSuppressions` not being applied on target repositories with `com.google.errorprone>=2.37.0`
==COMMIT_MSG==


CRogers:
>suppressible-error-prone is on errorprone 2.31.0, meaning that RemoveRolloutSuppressions ran despite it's severity being set to OFF erroneously, allowing the test to slyly pass.
>
>However, some of our other repositories are on e.g. errorprone 2.40.0. On those repositories, RemoveRolloutSuppressions didn't run (which is expected).

Oh this is nasty!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/ApplyMode.java
```diff
@@ -47,10 +46,6 @@ private static Set<String> checksToApplySuggestedPatchesFor(ModeOptionContext co
                     .collect(Collectors.toSet());
         }
 
-        return context.extension().patchChecksForCompilation(context.javaCompile()).stream()
-                // Do not patch checks that have been explicitly disabled
-                .filter(check ->
-                        context.errorProneOptions().getChecks().getting(check).getOrNull() != CheckSeverity.OFF)
-                .collect(Collectors.toSet());
```
CRogers, line 54: Is the point here we no longer need to do this as error-prone is now doing it for us?

kelvinou01: yep

CRogers (approved):

PR #160, opened 2025-09-24
Prevent adding suppressions to lambda implicit parameters and anonymous classes

aldexis:
## Before this PR
There is a bug currently where suppressible-error-prone considers lambda implicit parameters and anonymous classes as suppressible, which isn't actually the case.

## After this PR
==COMMIT_MSG==
Prevent adding suppressions to lambda implicit parameters and anonymous classes
==COMMIT_MSG==

This PR adds:
- A new submodule containing test-specific error-prone checks, so we can run tests with minimally reproducing checks (I couldn't find in error-prone that would do what I want, and decided I might as well do this, which can be beneficial for other tests too)
- Tests that verify we can apply the suppressions on the correct level (whether the lambda parameters are implicit or explicit)
- The actual bug fix


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -70,6 +70,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
             
             dependencies {
                 errorprone 'com.google.errorprone:error_prone_core:2.31.0'
+                // TODO: figure out a way to get the current version, rather than use the latest here
+                errorprone 'com.palantir.suppressible-error-prone:test-error-prone-checks:+'
```
aldexis: Taker of ideas as to how to do this. This works, but I'm mostly concerned about bugs where we somehow already have a higher version published locally (e.g. due to switching branches)

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -333,6 +335,52 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+                    Stream.of(new Object()).forEach((Object o) -> o.toString());
+                }
+            }
+        '''.stripIndent(true)
+
+        when:
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneSuppress')
+
+        then:
+        appJavaTextContains('@SuppressWarnings(\"for-rollout:TestCheckNoSingleLetterVariable\")\nObject o')
```
aldexis: So this always adds a newline, which leads to code looking like this, no matter the length of the check name (or so it seems)
```
    void test() {
        Stream.of(new Object()).forEach((@SuppressWarnings("for-rollout:TestCheckNoSingleLetterVariable")
Object o) -> o.toString());
    }
```
This seems to be geared towards field and variable declarations, which makes me wonder whether we should suppress here 🤔 (also curious what happens for e.g. method parameters)

Is the expectation that the formatter will then come and improve this? Do we even truly want suppressions at this level?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -135,8 +138,38 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+        if (!(lambdaExpressionTree instanceof JCLambda jcLambda)) {
+            return false;
+        }
+
+        if (jcLambda.paramKind != ParameterKind.IMPLICIT) {
+            return false;
+        }
+
+        return lambdaExpressionTree.getParameters().contains(variableTree);
     }
```
aldexis: This is the important part that actually fixes the bug. I tried to see whether there was anything on the `Tree` itself that could tell us whether the element can be suppressed, but there was no difference between e.g. implicit and explicit lambda parameters, beyond the `paramKind` itself, and the latter _is_ suppressible

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
esword: is this what you're looking for - https://github.com/palantir/gradle-plugin-testing?tab=readme-ov-file#resolution-of-dependencies-in-generated-build-files

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
crogoz: ugh, just ran the formatter on this code and it doesn't look too great  😢

>     void test() {
>         Stream.of(new Object())
>                 .forEach(
>                         (@SuppressWarnings("for-rollout:TestCheckNoSingleLetterVariable")
>                                 Object o) -> o.toString());
>     }

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -135,8 +138,50 @@ public static Description interceptDescription(VisitorState visitorState, Descri
                 .build();
     }
 
-    private static boolean suppressibleTree(Tree tree) {
-        return modifiersTree(tree).isPresent();
+    private static boolean suppressibleTreePath(TreePath treePath) {
+        if (!modifiersTree(treePath.getLeaf()).isPresent()) {
+            return false;
+        }
```
aldexis: At this point, I'm inclined to just make this a check for whether it's an instance of ClassTree, MethodTree or VariableTree. Being suppressible isn't actually tied to having a ModifiersTree it would seem

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -333,6 +335,79 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+            import java.util.stream.Stream;
+            public class App {
+                void test() {
+                    new Object() {
+                        {
+                            Stream.of(new Object()).forEach(o -> o.toString());
+                        }
+                    };
+                }
+            }
```
aldexis, line 421: This would previously add it as
```
public class App {
    void test() {
        new Object() @SuppressWarnings("for-rollout:TestCheckNoSingleLetterVariable")
 {
            {
                Stream.of(new Object()).forEach(o -> o.toString());
            }
        };
    }
}
```

which simply isn't valid

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
aldexis: Oh right, I'll have a look at whether this is something I could leverage here!

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
aldexis: What if it's `@SuppressWarnings("for-rollout:deprecation")`? (since that's what we're likely to hit)
I agree it's not great either case. Wondering whether @CRogers has some thoughts here. I wouldn't mind considering variable declarations as non-suppressible, except for fields

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
aldexis: I think this is something we can use for getting the versions from the versions.props file, but I'm not sure this is possible to have it use the "current project version"? In any case, I've found a work-around by relying on the same gradle property that the plugin uses https://github.com/palantir/suppressible-error-prone/blob/a141dc7eb4d5c198a4203184f58fceed617e7516/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L69

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: oof this is nasty one

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: I really like the suppressions on local variables in methods. It makes it far more obvious what the suppression applies to. 

I don't think we should apply suppressions onto lambda parameters, implicit or explicit. I don't think it is clear what this means (it looks like the suppressions are being applied to the lambda parameter only rather than whole lambda body too). It's also extremely ugly.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -135,8 +138,50 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+
+        if (!(tree.getParentPath().getLeaf() instanceof LambdaExpressionTree lambdaExpressionTree)) {
+            return false;
+        }
+
+        if (!(lambdaExpressionTree instanceof JCLambda jcLambda)) {
+            return false;
+        }
+
+        if (jcLambda.paramKind != ParameterKind.IMPLICIT) {
```
CRogers: As I said in the other comment, I'd just treat explicit lambda params as unsuppressible too.

test-error-prone-checks/src/main/java/com/palantir/gradle/suppressibleerrorprone/TestCheckNoSingleLetterVariable.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "This check is meant only for testing suppressible-error-prone functionality")
+public class TestCheckNoSingleLetterVariable extends BugChecker implements VariableTreeMatcher {
```
CRogers, line 32: Nice, really I should have made custom checks for everything in the tests rather than depend on the built in ones.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
esword: Ah ya - That's this part:
https://github.com/palantir/gradle-plugin-testing/blob/develop/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java#L127-L147

https://github.com/palantir/gradle-plugin-testing/blob/develop/gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/AbstractTestingPluginSpec.groovy#L26-L27

https://github.com/palantir/suppressible-error-prone/blob/a141dc7eb4d5c198a4203184f58fceed617e7516/gradle-suppressible-error-prone/build.gradle#L40

Which is a technique Callum told me about that we do in a few places like here. Makes me wonder if the PluginTestingPlugin should somehow provide utilities to do this as well so we stop copy-pasting this technique everywhere.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
esword: Are you saying we should not suppress it and let it error or that we should put the suppression on the next higher level up so it looks cleaner?  This relates to an actual case where PR is failing that I referred Alexis to from a thread in our team room.  I'll tag you on the thread for easier reference.

aldexis: Pretty sure this means applying it to a higher level (by indicating that "this" is not suppressible).

@CRogers What's your opinion on method arguments? Seems like we've discovered a case (see internal thread) where it doesn't work for at least NullAway. And I generally think this is ugly and doesn't provide much more benefits than adding it to the method 🤔

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
aldexis: @CRogers This should be done now. I haven't made any change to method parameters for now and can do so in a subsequent PR if we think it's what we want

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -135,8 +136,44 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+    private static boolean isLambdaParameter(TreePath tree) {
+        if (!(tree.getLeaf() instanceof VariableTree variableTree)) {
+            return false;
+        }
+
+        if (!(tree.getParentPath().getLeaf() instanceof LambdaExpressionTree lambdaExpressionTree)) {
+            return false;
+        }
+
+        return lambdaExpressionTree.getParameters().contains(variableTree);
```
kelvinou01: Nice! Let's add this to `SuppressWarningsUtils`? I have a use case for this in a future PR : )

aldexis: Good idea! Done!

CRogers (approved):

PR #163, opened 2025-09-26
fix annotation on anonymous class method param

esword:
## Before this PR
Various excavator PRs that marked FutureCallback with NullAway are failing because the suppression isn't in the right place.  Maybe.

## After this PR
This PR doesn't fix the problem yet.  Only adds a test which shows the issue

==COMMIT_MSG==
==COMMIT_MSG==


CRogers:
Superceeded by #160

aldexis:
@CRogers I don't think it is actually. I'm pretty sure this test still doesn't pass, because _method_ parameters are still considered suppressible. Happy to update this if we think it's reasonable

PR #166, opened 2025-09-29
Add workaround for debugging errorprones in `SuppressibleErrorPronePluginIntegrationTest`

kelvinou01:
## Before this PR

The mechanism used to debug errorprones started failing with the following message

```
ERROR: transport error 202: connect failed: Connection refused
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:700]
```

The root cause was that we started running tests with configuration cache. This doesn't play well with the JWDP for some reason. 

## After this PR

Move `debuggingErrorPrones` to a top-level variable, and don't run configuration cache when we are debugging.

==COMMIT_MSG==
Add workaround for debugging errorprones in `SuppressibleErrorPronePluginIntegrationTest`
==COMMIT_MSG==

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -40,7 +40,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
     //   5. Run the tests as well
     // If the variable below is true the tests will fail as the compilation process will try to
     // attach to a non-existent debugger. Set it to false before you push any code.
-    boolean debuggingErrorPrones = true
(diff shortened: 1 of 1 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -30,6 +30,18 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
     File mainSourceSet
     File otherSourceSet
 
+    // This makes debugging the errorprone check code running inside the compiler (including the bytecode
```
felixdesouza, line 33: as a flup, we should make gradle generate a run configuration that sets up this debugger for the debugged build

felixdesouza:
👍

PR #167, opened 2025-09-29
-PerrorProneRemoveUnused implementation (works with suppress and apply): remove unused suppressions on every encountered compilation unit in VisitorStateModifications

kelvinou01:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
```diff
@@ -27,42 +27,37 @@
...
+ * A Replacement that handles @SuppressWarnings modifications with line-removing capabilities.
+ * When the final replacement text is empty (no suppressions), it will remove preceding
+ * whitespace up to and including the newline.
+ */
+final class LazySuppressionReplacement extends Replacement {
     private final Range<Integer> range;
     private final List<String> existingSuppressions;
     private final String suffix;
-    private final Set<String> newSuppressions;
+    private final Set<String> desiredSuppressions;
```
kelvinou01, line 39: This implicitly uses a listener/accumulator pattern, like error-prone's DescriptionListener. Maybe make this explicit

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtils.java
```diff
@@ -74,5 +90,23 @@ public static String suppressWarningsString(List<String> warningsToSuppress) {
         return "@" + CommonConstants.SUPPRESS_WARNINGS_ANNOTATION + "(" + suppressWarningsString + ")";
     }
 
+    /**
+     * Only these trees are suppressible, as per
```
kelvinou01: This cond is too loose! See #160

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1313,6 +1313,336 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+                    
+                    static class InnerInner {
+                        void method() {
+                            Arrays.equals(new int[3], new int[3]);
+                        } 
+                    }
+                }
+            }
+    '''.stripIndent(true)
+    }
```
kelvinou01, line 1938: ## 🚨 TLDR of what this PR enables

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/SuppressibleTreePathScannerClassVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * <p>{@code RemoveUnusedSuppressions} must see violations that are normally hidden by suppressions to
+ * determine which suppressions are actually needed. While ErrorProneOptions has an
+ * {@code ignoreSuppressionAnnotations} flag for this purpose, many BugCheckers use
+ * {@code SuppressibleTreePathScanner}, which doesn't respect this flag.
+ *
+ * <p>This class uses bytecode manipulation to patch {@code SuppressibleTreePathScanner::isSuppressed}
+ * to respect the ErrorProneOptions setting.
+ */
+final class SuppressibleTreePathScannerClassVisitor extends ClassVisitor {
```
CRogers, line 35: As discussed in person, we should try to get the issue fixed in error-prone before resorting to further asm modification.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
CRogers: In that it's a mutable set?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtils.java
CRogers: We need to be careful here. We've decided that suppressible-error-prone will not put suppressions in certain locations, but if we're removing unused human suppressions they may be in these locations. This may complicate things somewhat.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -23,35 +23,91 @@
...
 
-    @SuppressWarnings("RestrictedApi")
+    @SuppressWarnings("CyclomaticComplexity") // Unfortunately this logic is just complex
     public static Description interceptDescription(VisitorState visitorState, Description description) {
-        if (description == Description.NO_MATCH) {
+        if (description == Description.NO_MATCH || description.checkName.equals("SuppressibleErrorProne")) {
             return description;
         }
 
+        Set<String> modes = visitorState.errorProneOptions().getFlags().getSetOrEmpty("SuppressibleErrorProne:Mode");
```
CRogers, line 102: ~~Mmm, putting the mode in here kinda odd, especially as we could be in multiple modes.~~ just realised this is a Set, but maybe we could have a nice abstract here - would need to think about this more.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtils.java
CRogers: Especially as we might want to "move" `for-rollout:` suppressions to be in a more specific location (eg because we changed up some errorprone logic), but we probably don't want to do this to human authored suppressions unless we need to (eg someone suppresses a whole class to stop loads of errors - we probably want to keep this).

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/SuppressibleTreePathScannerClassVisitor.java
kelvinou01: yep, lets see how the PR does

https://github.com/google/error-prone/pull/5255

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressWarningsUtils.java
kelvinou01: > we probably don't want to do this to human authored suppressions unless we need to

yep, RemoveUnusedMode has [tests for this specific case](https://github.com/palantir/suppressible-error-prone/pull/167/files#diff-c7534f7be5e7afc86368f21228579a5a830def655d7b3eecf842e3ce0c504cffR1434)

PR #172, opened 2025-09-30
Fix apply + remove rollout overlapping with one another

kelvinou01:
## Before this PR

Resolves https://github.com/palantir/suppressible-error-prone/pull/70. 

Our whitespace formatting in RemoveRollout's LineRemovingReplacementFix was overlapping with other fixes, causing LineRemovingReplacementFix to be dropped. 

## After this PR

- Instead of removing whitespace and newlines by expanding the start of the fix, we do it by expanding the end of the fix. RemoveRollout now isn't dropped when there is a fix on the same AnnotationTree. 
- Use the formatter in our integration tests so tests still can pass. This is okay as suppressible-error-prone is always run with the formatter both in local and in excavators anyways

==COMMIT_MSG==
Fix apply + remove rollout overlapping with one another
==COMMIT_MSG==

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -18,14 +18,18 @@ package com.palantir.gradle.suppressibleerrorprone
 
 
 import com.palantir.gradle.plugintesting.ConfigurationCacheSpec
+import com.palantir.javaformat.java.JavaFormatterOptions
 import org.apache.commons.io.FileUtils
 import org.gradle.testkit.runner.BuildResult
 import spock.lang.Unroll
+import com.palantir.javaformat.java.Formatter
+
 
 class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec {
     // We need to put the source sets in a different directory that does not contain the any words that would hit
     // the errorprone excludedPathRegex, ie build in build/nebulatest
     static File nebulatestSourceSets = new File('nebulatestSourceSets/' + SuppressibleErrorPronePluginIntegrationTest.class.simpleName)
+    static Formatter formatter = Formatter.createFormatter(JavaFormatterOptions.builder().style(JavaFormatterOptions.Style.PALANTIR).build())
     File sourceSetRoot
     File mainSourceSet
     File otherSourceSet
@@ -40,7 +44,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
     //   5. Run the tests as well
     // If the variable below is true the tests will fail as the compilation process will try to
     // attach to a non-existent debugger. Set it to false before you push any code.
-    boolean debuggingErrorPrones = isJwdpLoaded()
+    boolean debuggingErrorPrones = false
 
     def setupSpec() {
         FileUtils.deleteDirectory(nebulatestSourceSets)
@@ -94,7 +98,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     // These interfere with some tests, so disable them
                     // TODO(callumr): Rewrite the tests to use custom testing error-prones rather than built in checks
                     //                to make upgrading error-prone easier.
-                    disable('Varifier', 'ReturnValueIgnored', 'UnusedVariable', 'IdentifierName')
+                    disable('Varifier', 'ReturnValueIgnored', 'UnusedVariable', 'IdentifierName', 'UnusedMethod')
                     ignoreUnknownCheckNames = true
                 }
             }
@@ -112,7 +116,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     it.options.forkOptions.jvmArgumentProviders.add(new CommandLineArgumentProvider() {
                         @Override
                         public Iterable<String> asArguments() {
-                            return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5006")
+                            return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5005")
                         }
                     })
                 }
@@ -131,6 +135,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
@@ -153,6 +158,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 @SuppressWarnings("for-rollout:ArrayToString")
                 public static void main(String[] args) {
@@ -169,6 +175,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         def erroringCode = '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
@@ -203,6 +210,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
@@ -216,7 +224,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         then:
         runTasksSuccessfully('compileAllErrorProne')
 
-        appJavaTextContains('Arrays.toString(new int[3])')
+        javaSourceContains('Arrays.toString(new int[3])')
     }
 
     def 'does not apply patches for a check if not added to the patchChecks list'() {
@@ -231,6 +239,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
@@ -242,7 +251,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         runTasksSuccessfully('compileAllErrorProne', '-PerrorProneApply')
 
         then:
-        appJavaTextContains('new int[3].toString()')
+        javaSourceContains('new int[3].toString()')
     }
 
     def 'does not apply patches if there is nothing in patchChecks set'() {
@@ -256,6 +265,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
@@ -269,7 +279,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
 
         then:
         stderr.contains('[ArrayToString]')
-        appJavaTextContains('new int[3].toString()')
+        javaSourceContains('new int[3].toString()')
     }
 
     def 'does not apply patches for check that was explicitly disabled'() {
@@ -287,6 +297,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         // language=Java
         writeJavaSourceFileToSourceSets '''
             package app;
+            
             public final class App {
                 public static void main(String[] args) {
                     new int[3].toString();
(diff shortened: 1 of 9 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1484,23 +1516,44 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+    void resultContains(String substring) {
         assert file('app/App.java', mainSourceSet).text.contains(substring)
         assert file('app/App.java', otherSourceSet).text.contains(substring)
     }
 
-    void appJavaTextNotContains(String substring) {
+    void resultDoesNotContain(String substring) {
         assert !file('app/App.java', mainSourceSet).text.contains(substring)
         assert !file('app/App.java', otherSourceSet).text.contains(substring)
     }
```
kelvinou01: Renamed these to be consistent with `resultIsSyntacticallyEqualTo`. And `appJavaIsSyntacticallyEqualTo` didn't sound quite right to me

CRogers: I would actually keep the previous naming - `result` is very generic to me. Maybe we should call it `javaSourceIsSyntacticallyEqualTo` to match up with `writeJavaSourceFileToSourceSets`

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1484,23 +1516,44 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+        String stripped = content.readLines()
+                .collect { it.trim() }           // Remove leading/trailing whitespace
+                .findAll { !it.isEmpty() }       // Remove empty lines
+                .join('\n')
+        return Formatter.create().formatSource(stripped)
+    }
+
+    void resultIsSyntacticallyEqualTo(String source) {
+        var output = normalizeSource(file('app/App.java', mainSourceSet).text)
+        var expected = normalizeSource(source)
```
CRogers: We should assert the expected value in the string literal is correct

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1206,6 +1208,36 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+                @SuppressWarnings("for-rollout:ShouldBeNullable")
+                public Object iAmNullable() {
+                    System.out.println("Test");
+                    return null;
+                }
+            }
+        '''.stripIndent(true)
+        when:
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveRollout=ShouldBeNullable', '-PerrorProneApply=ShouldBeNullable')
+        then:
```
CRogers, line 1302: newline above then: and when:

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -89,77 +83,7 @@ public Description matchAnnotation(AnnotationTree tree, VisitorState state) {
 
         String updatedText = SuppressWarningsUtils.suppressWarningsString(updatedSuppressions);
 
-        return buildDescription(tree)
-                .addFix(new LineRemovingReplacementFix(state.getSourceCode(), (DiagnosticPosition) tree, updatedText))
```
CRogers, line 93: Let's add a bit to the readme saying that we will not output nicely formatted code after this, you must use a formatter.

CRogers (commented):
LGTM except those comments

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
kelvinou01: Yep, that makes sense! Ill use javaSource* for everything.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
kelvinou01: Good catch

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
kelvinou01: Done

CRogers:
As discussed in person, this causes issues with leaving a blank line when removing a suppress warnings annotation. palantir-java-format allows users to specify up to one blank line, so this can result in extra blank lines being added if you remove all suppressions and add them back again. Instead we're going to investigate removing trailing rather than preceding whitespace, which we suspect won't have the same overlapping problem as bugcheckers usually insert at the start of element spans, not at the end.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -40,7 +40,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
     //   5. Run the tests as well
     // If the variable below is true the tests will fail as the compilation process will try to
     // attach to a non-existent debugger. Set it to false before you push any code.
-    boolean debuggingErrorPrones = false
+    boolean debuggingErrorPrones = isJwdpLoaded()
```
kelvinou01: ignore these changes, added for testing

kelvinou01:
As demonstrated in [`1086496`](https://github.com/palantir/suppressible-error-prone/commit/10864962daabfb425982f729bab4695e7db363fa), removing trailing rather than preceding whitespace also leads to overlapping with other fixes, particularly fixes done on visibility modifiers of functions.

kelvinou01:
> removing trailing rather than preceding whitespace also leads to overlapping with other fixes, particularly fixes done on visibility modifiers of functions.

as discussed privately, lets expand the end of the fix up until the newline, but not including the whitespace. leave the whitespace for the formatter. this is kinda nasty but does the trick for us

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1206,10 +1274,91 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+                
+                @Nullable
+                private Integer fixme(Integer i) {
+                    return null;
+                }
+            }
+        '''.stripIndent(true)
+    }
+
+    def '-PerrorProneSuppress then -PerrorProneRemoveRollout does not add newlines'() {
```
CRogers, line 1328: nice

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
```diff
@@ -139,9 +139,8 @@ public ImmutableSet<Replacement> getReplacements(EndPosTable endPositions) {
             // If we are looking to delete the entire element, we should also remove whitespace before it,
             //   up to and including the newline
             if (replacementText.isEmpty() && sourceCode != null) {
-                int start = SourceCodeUtils.startPositionWithWhitespaceIncludingNewLine(
```
CRogers, line 142: Can we remove `startPositionWithWhitespaceIncludingNewLine` and it's tests now?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
```diff
@@ -63,5 +63,26 @@ private static int startPositionWithWhitespace(CharSequence sourceCode, int sour
         return pos + 1;
     }
 
+    static int removeUntil(CharSequence sourceCode, int endOfSuppression) {
```
CRogers: This isn't actually removing anything right, just giving back an index? And it's stopping at the first newline?  Perhaps call this `whitespaceUpToFirstNewline`?

Similarly, I'd call `endOfSuppression` `startPosition` instead - this method doesn't need to be specific to suppressions.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
```diff
@@ -63,5 +63,26 @@ private static int startPositionWithWhitespace(CharSequence sourceCode, int sour
...
+            char character = sourceCode.charAt(pos);
+            if (character == '\n' || !Character.isWhitespace(character)) {
+                break;
+            }
+        }
+
+        if (pos < sourceCode.length() && sourceCode.charAt(pos) == '\n') {
+            // If the next element is in a newline, remove the newline as well.
+            // Ideally, we also remove whitespace before the next element, but that would overlap with any done on the
+            // next element. We leave those spaces to the formatter.
```
CRogers: I'd leave this comment in the `LineRemovingReplacementFix`. It's more relevant there than this generic utility function.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
```diff
@@ -63,5 +63,26 @@ private static int startPositionWithWhitespace(CharSequence sourceCode, int sour
...
+            return pos + 1;
```
CRogers: Can we just do this where the `break` in the `for` loop currently is? Then avoid this latest if-else entirely and just `return pos`?

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1484,23 +1646,50 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+    private static String normalizeSource(String content) {
+        String stripped = content.readLines()
+                .collect { it.trim() }           // Remove leading/trailing whitespace
+                .join('\n')
+
+        return formatter.formatSource(stripped)
+    }
+
+    void javaSourceIsSyntacticallyEqualTo(String source) {
+        var output = normalizeSource(file('app/App.java', mainSourceSet).text)
```
CRogers: Generally we use `def` rather than `var` in groovy code. It's a bit weird seeing Java `var` here.

CRogers (commented):
LGTM apart from comments

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
kelvinou01: Yep!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
kelvinou01: yep, how about `firstNonWhitespaceOrNextLineStart`?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SourceCodeUtils.java
kelvinou01: Yep, I've simplified the logic, let me know what you think!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/RemoveRolloutSuppressions.java
kelvinou01: Sadly no, it's still being used by SuppressingFix

CRogers: We're we going to get SuppressingFix to use this as well? Ah, it was for the remove unused suppresisons? Are we going to break that out into another smaller PR before the big unused suppressions one?

CRogers (approved):

PR #181, opened 2025-10-03
Create the `JdwpRemoteDebugPlugin` and the `Debug errorprones` Intellij debug config, to make debugging error-prones easier

kelvinou01:
## Before this PR

In this repo, and every repo which consumes suppressible-error-prone, to debug errorprones, we have to 

1. Create a Remote JVM Debug config following some instructions
2. Add the snippet below to the build-under-test
3. Run the debug config
4. Run the test

```java
tasks.withType(JavaCompile).configureEach {
    it.options.forkOptions.jvmArgumentProviders.add(new CommandLineArgumentProvider() {
        @Override
            return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5006")
        }
    })
}
```

## After this PR

Simplify step 1 by creating a debug config file, which can also be replicated to other repositories like gradle-baseline

Also, simplify step 2 by exporting a gradle plugin which does the above for you.

==COMMIT_MSG==
Create the `JdwpRemoteDebugPlugin` and the `Debug errorprones` Intellij debug config, to make debugging error-prones easier
==COMMIT_MSG==

## Possible downsides?

Removing step 1 would be a game changer. I tried to do this with a [compound task](https://www.jetbrains.com/help/idea/run-debug-multiple.html) which ran both the jdwp listener and the test build, but there was no way to run a compound task which runs a specific test case, like how intellij allows you to run test cases from the gutter.

### DIFF
--- a/.idea/runConfigurations/Debug_errorprones.xml
+++ b/.idea/runConfigurations/Debug_errorprones.xml
@@ -1,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
+    <configuration default="false" name="Debug errorprones" type="Remote">
+        <option name="USE_SOCKET_TRANSPORT" value="true" />
+        <option name="SERVER_MODE" value="true" />
+        <option name="SHMEM_ADDRESS" />
+        <option name="HOST" value="localhost" />
+        <option name="PORT" value="5006" />
+        <option name="AUTO_RESTART" value="true" />
+        <RunnerSettings RunnerId="Debug">
+            <option name="DEBUG_PORT" value="5006" />
+            <option name="LOCAL" value="false" />
+        </RunnerSettings>
+        <method v="2" />
+    </configuration>
+</component>
--- a/gradle-suppressible-error-prone/build.gradle
+++ b/gradle-suppressible-error-prone/build.gradle
@@ -33,11 +33,11 @@ gradlePlugin {
             implementationClass = 'com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin'
         }
         jdwpRemoteDebugPlugin {
-            id = 'com.palantir.jdwp-remote-debug'
+            id = 'com.palantir.remote-debug-java-compile'
             displayName = 'JDWP Remote Debug Plugin'
             description = 'Configures Java compilation tasks to connect to a remote JDWP debugger for debugging ErrorProne and other compiler processes.'
             tags = ['jdwp', 'remote-debug', 'java', 'debugging', 'compiler']
-            implementationClass = 'com.palantir.gradle.suppressibleerrorprone.JdwpRemoteDebugPlugin'
+            implementationClass = 'com.palantir.gradle.suppressibleerrorprone.RemoteDebugJavaCompilePlugin'
         }
     }
 }
--- a/gradle-suppressible-error-prone/build.gradle
+++ b/gradle-suppressible-error-prone/build.gradle
@@ -32,6 +32,13 @@ gradlePlugin {
             tags = ['error-prone', 'javac', 'suppression']
             implementationClass = 'com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin'
         }
+        jdwpRemoteDebugPlugin {
+            id = 'com.palantir.jdwp-remote-debug'
+            displayName = 'JDWP Remote Debug Plugin'
+            description = 'Configures Java compilation tasks to connect to a remote JDWP debugger for debugging ErrorProne and other compiler processes.'
+            tags = ['jdwp', 'remote-debug', 'java', 'debugging', 'compiler']
+            implementationClass = 'com.palantir.gradle.suppressibleerrorprone.JdwpRemoteDebugPlugin'
+        }
     }
 }
 
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
@@ -24,10 +24,9 @@ import org.gradle.process.CommandLineArgumentProvider;
 
 /**
  * To debug suppressible-error-prone in a build (works with breakpoints in VisitorStateModifications):
- * 1. Apply this plugin to the build
- * 2. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
+ * 1. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
  *    at port 5006
- * 3. Run the build
+ * 2. Run the build with this plugin applied
  */
 public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
     // No other way to programmatically check whether the configuration-cache is on
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
@@ -25,8 +25,8 @@ import org.gradle.process.CommandLineArgumentProvider;
 /**
  * To debug suppressible-error-prone in a build (works with breakpoints in VisitorStateModifications):
  * 1. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
- *    at port 5006
- * 2. Run the build with this plugin applied
+ *    on port 5006
+ * 2. Run the build-under-test with this plugin applied
  */
 public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
     @Inject
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
@@ -17,8 +17,10 @@
 package com.palantir.gradle.suppressibleerrorprone;
 
 import java.util.List;
+import javax.inject.Inject;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.configuration.BuildFeatures;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.process.CommandLineArgumentProvider;
 
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
@@ -1,0 +1,60 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.suppressibleerrorprone;
+
+import java.util.List;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.compile.JavaCompile;
+import org.gradle.process.CommandLineArgumentProvider;
+
+/**
+ * To debug suppressible-error-prone in a build (works with breakpoints in VisitorStateModifications):
+ * 1. Apply this plugin to the build
+ * 2. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
+ *    at port 5006
+ * 3. Run the build
+ */
+public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
+    // No other way to programmatically check whether the configuration-cache is on
+    @SuppressWarnings("DeprecatedApiUsage")
+    @Override
+    public final void apply(Project project) {
+
+        if (project.getGradle().getStartParameter().isConfigurationCacheRequested()) {
+            throw new IllegalArgumentException(
+                    "The JDWP will throw a cryptic error when run with the configuration cache. Turn off configuration"
+                            + " cache for the build-under-debug. Hint: you can conditionally apply "
+                            + "`com.palantir.jdwp-remote-debug` only if `IntegrationTestKitBase#isJdwpLoaded()`");
+        }
+
+        project.getPluginManager().withPlugin("java", unused -> {
+            project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
+                javaCompile
+                        .getOptions()
+                        .getForkOptions()
+                        .getJvmArgumentProviders()
+                        .add(new CommandLineArgumentProvider() {
+                            @Override
+                            public Iterable<String> asArguments() {
+                                return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5006");
+                            }
+                        });
+            });
+        });
+    }
+}
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
@@ -29,12 +29,12 @@ import org.gradle.process.CommandLineArgumentProvider;
  * 2. Run the build with this plugin applied
  */
 public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
-    // No other way to programmatically check whether the configuration-cache is on
-    @SuppressWarnings("DeprecatedApiUsage")
+    @Inject
+    protected abstract BuildFeatures getBuildFeatures();
+
     @Override
     public final void apply(Project project) {
-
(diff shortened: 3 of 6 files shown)


.idea/runConfigurations/Debug_errorprones.xml
```diff
@@ -0,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
```
aldexis, line 1: Is this intentional? I've generally seen run configurations be added by leveraging the idea gradle plugin

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/RemoteDebugJavaCompilePlugin.java
```diff
@@ -0,0 +1,59 @@
...
+                        .getForkOptions()
+                        .getJvmArgumentProviders()
+                        .add(new CommandLineArgumentProvider() {
+                            @Override
+                            public Iterable<String> asArguments() {
+                                return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5006");
+                            }
+                        });
+            });
+        });
```
aldexis, line 60: Whenever I've debugged, I've wanted to debug a specific compilation task. If it requires to run another one it depends on, wouldn't this cause me to listen to the wrong one, since it's adding this to all the compilation tasks?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+    // No other way to programmatically check whether the configuration-cache is on
+    @SuppressWarnings("DeprecatedApiUsage")
+    @Override
+    public final void apply(Project project) {
+
+        if (project.getGradle().getStartParameter().isConfigurationCacheRequested()) {
+            throw new IllegalArgumentException(
+                    "The JDWP will throw a cryptic error when run with the configuration cache. Turn off configuration"
+                            + " cache for the build-under-debug. Hint: you can conditionally apply "
+                            + "`com.palantir.jdwp-remote-debug` only if `IntegrationTestKitBase#isJdwpLoaded()`");
```
aldexis: I personally don't know off the top of my head how to disable the configuration cache. It might be good to provide instructions in here as to how to do this? (or even just forcefully disable it?)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+import org.gradle.api.tasks.compile.JavaCompile;
+import org.gradle.process.CommandLineArgumentProvider;
+
+/**
+ * To debug suppressible-error-prone in a build (works with breakpoints in VisitorStateModifications):
+ * 1. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
+ *    at port 5006
+ * 2. Run the build with this plugin applied
+ */
+public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
```
felixdesouza: let's move this to the test source set

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.compile.JavaCompile;
+import org.gradle.process.CommandLineArgumentProvider;
+
+/**
+ * To debug suppressible-error-prone in a build (works with breakpoints in VisitorStateModifications):
+ * 1. Run the "Debug errorprones" debug config which ships with this repository. Notably, it sets up a jdwp listener
+ *    at port 5006
```
felixdesouza:
```suggestion
 *    on port 5006
```

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+ * 2. Run the build with this plugin applied
```
felixdesouza: be specific about what build you're referring to i.e. the build the under test

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public abstract class JdwpRemoteDebugPlugin implements Plugin<Project> {
+    // No other way to programmatically check whether the configuration-cache is on
```
felixdesouza: not true! 

Use `BuildFeatures#getConfigurationCache`

felixdesouza: https://docs.gradle.org/current/javadoc/org/gradle/api/configuration/BuildFeatures.html

gradle-suppressible-error-prone/build.gradle
```diff
@@ -32,6 +32,13 @@ gradlePlugin {
             tags = ['error-prone', 'javac', 'suppression']
             implementationClass = 'com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin'
         }
+        jdwpRemoteDebugPlugin {
+            id = 'com.palantir.jdwp-remote-debug'
+            displayName = 'JDWP Remote Debug Plugin'
+            description = 'Configures Java compilation tasks to connect to a remote JDWP debugger for debugging ErrorProne and other compiler processes.'
+            tags = ['jdwp', 'remote-debug', 'java', 'debugging', 'compiler']
+            implementationClass = 'com.palantir.gradle.suppressibleerrorprone.JdwpRemoteDebugPlugin'
+        }
```
felixdesouza, line 41: is this something you want to publish? or you only want it for your tests. If the latter, we should not publish this here.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -105,17 +94,10 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
             buildFile << '''
-                tasks.withType(JavaCompile).configureEach {
-                    it.options.forkOptions.jvmArgumentProviders.add(new CommandLineArgumentProvider() {
-                        @Override
-                        public Iterable<String> asArguments() {
-                            return List.of("-agentlib:jdwp=transport=dt_socket,server=n,address=localhost:5005")
-                        }
-                    })
-                }
+                apply plugin: 'com.palantir.jdwp-remote-debug'
```
felixdesouza: I like that you've encapsulated this in a plugin, pretty neat

gradle-suppressible-error-prone/build.gradle
felixdesouza: If the former, then it's probably a bit weird to publish *within* `gradle-suppressible-error-prone`

.idea/runConfigurations/Debug_errorprones.xml
kelvinou01: yep! im using https://www.jetbrains.com/help/idea/run-debug-configuration.html#share-configurations

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/RemoteDebugJavaCompilePlugin.java
kelvinou01: hmm, i thought you'd only be running one compilation task, no? whats an example where there are multiple compilations being done?

gradle-suppressible-error-prone/build.gradle
kelvinou01: We want to publish this. @aldexis wants to use it in e.g baseline as well

gradle-suppressible-error-prone/build.gradle
```diff
@@ -32,6 +32,13 @@ gradlePlugin {
             tags = ['error-prone', 'javac', 'suppression']
             implementationClass = 'com.palantir.gradle.suppressibleerrorprone.SuppressibleErrorPronePlugin'
         }
+        jdwpRemoteDebugPlugin {
+            id = 'com.palantir.jdwp-remote-debug'
```
CRogers: This id is too general. I'd call it `com.palantir.remote-debug-java-compiler`

gradle-suppressible-error-prone/build.gradle
felixdesouza: if that's the case, then we should move it to [gradle-utils](https://github.com/palantir/gradle-utils), currently, you need to depend on `gradle-suppressible-errorprone` to get this, which is weird.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -30,23 +30,16 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
 
     def setupSpec() {
         FileUtils.deleteDirectory(nebulatestSourceSets)
     }
 
     def setup() {
+        // To debug the tests in this file
+        // 1. Run the "Debug errorprones" debug configuration, which ships with this repository
+        // 2. Go to the test you're looking for and run debug on it as well, but without any special configurations
+        debug = isJwdpLoaded()
```
CRogers, line 45: [Pretty sure that nebula-test's `IntegrationTestKitSpec` already does this](https://github.com/nebula-plugins/nebula-test/blob/7f9b9ba8bcbe0dd4f9f0acaf48b519315a9526b6/src/main/groovy/nebula/test/IntegrationTestKitBase.groovy#L107)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/RemoteDebugJavaCompilePlugin.java
felixdesouza: not necessarily. It'd be the case if you've set up the build-under-debug with multiple projects that compiling java, or a single project with multiple `JavaCompile` tasks, you can get this if you have different source sets that need compilation.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
kelvinou01: It's only on if you run with --configuration-cache (e.g. using the methods in `ConfigurationCacheSpec`), and I don't think there's a way to disable it once your build is running :/ would be nice if we could!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
kelvinou01: If we want to export this to other repos, e.g. gradle-baseline, should we have it in the test source set still?

felixdesouza: ignore, if you're intending to publish this

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
kelvinou01: yep i should make all my terminologies very explicit. thanks for raising this.

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
felixdesouza: yeah, but it sets it only when you're trying to create a runner, so your test code can't depend on it as it runs too late. It's kinda shit that they set it inside `createRunner`

.idea/runConfigurations/Debug_errorprones.xml
CRogers: This is fine for having it in just this repo - however, the real dream for a plugin like this would be you can use it debug errorprone sources too (like gradle-baseline) and get this in the error-prone source repo too automatically. I think you'd need to make another plugin for this though that you apply in the errorprone source (that adds this run configuration). I don't think we currently have a plugin like that atm though.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
kelvinou01: wow, thank you 🤯

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/JdwpRemoteDebugPlugin.java
CRogers: `--no-configuration-cache` - but yes this should be in error message

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: Ah, you mean if you want to do something different depending on if `debug` is set.

gradle-suppressible-error-prone/build.gradle
aldexis: To be honest, I would even prefer for this to just be a configuration option in suppressible-error-prone that I can just turn on

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
kelvinou01: It does, but only at runner creation time, which is too late unfortunately. we need it at setup time, when we create the build script

kelvinou01: so we have to set it here early. i should explain this in a comment.

gradle-suppressible-error-prone/build.gradle
kelvinou01: hmm we only use this debugging plugin when we also have suppressible-error-prone, is it general enough to be put in gradle-utils?

CRogers: I think the point is you can just slap this plugin on a repo using suppressible-error-prone without having to add another dependency to your buildscript - not really to be used as a generic compiler debug plugin used by other plugins.

felixdesouza: okay, that makes a bit more sense. Let's add this to the `README` as well pls 🙏

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/RemoteDebugJavaCompilePlugin.java
kelvinou01: We expect the user to apply the debug plugin to the correct subproject, and hence configure the correct javacompile task. is this reasonable?

aldexis: Not _quite_, because I might want to debug the test source set compilation, which typically depends on the main one. It should be enough in _most_ cases though

felixdesouza: if it is going to be used in other repos and not for the tests in suppressible errorprone, then it *is* confusing if you apply it in a repo that has more than one source set because you will have `compileJava`, `compileTestJava`, `compileIntegrationTestJava`, `compile<SourceSet>Java`. if you debug anything that's not the main source set, you may get into something weird.

My previous statement when you were devving on this was: it's fine to have a static file, but I assumed incorrectly on how it was supposed to be used. I think this would actually be perfect for having the dynamic pieces:
for each project that this is applied:
  1. have a unique port for each compile task
  2. generate a run configuration per project per compile task

so your menu would look like:
* `debug compile for gradle-suppressible-errorprone:compileJava`
* `debug compile for gradle-suppressible-errorprone:compileTestJava`

Not pushing for it in this PR, but it's a thought if you want this to be used widely

felixdesouza: @CRogers has enlightened me that it's in listen mode and its completely fine to have multiple things listen on the same port! ignore me!

.idea/runConfigurations/Debug_errorprones.xml
kelvinou01: Do you mean that we can add some code in RemoteDebugJavaCompilePlugin to create this configuration file in the repo we are running in, e.g. gradle-baseline, if it doesn't exist yet?

PR #192, opened 2025-10-13
Create a `LazySuppressionFix` to allow for lazily adding/removing suppressions

kelvinou01:
## Before this PR

We are working on `-PerrorProneRemoveUnused`, a mode which removes all unused suppressions. This mode requires us to lazily remove items from a `@SuppressWarnings` fix, and defer rendering the replacement until after all BugCheckers have been evaluated.

We already have a `SuppressingFix` which does the lazy evaluation when adding suppressions, and `LineRemovingReplacementFix` which handles removing suppression with proper line handling. Lets combine the functionality of both.

## After this PR

==COMMIT_MSG==
Create a `LazySuppressionFix` to allow for lazily adding/removing suppressions
==COMMIT_MSG==

### DIFF
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
@@ -33,6 +33,10 @@ import java.util.stream.Collectors;
  * whitespace up to and including the newline.
  */
 final class LazySuppressionReplacement extends Replacement {
+    // We really do need to be this lazy for generating the Replacements, as error-prone immediately converts
+    // the Fix to a Replacement when a Description is given to it, and we need to defer the computation of the
+    // Replacement until a number of Descriptions have been produced, to handle multiple errors being suppressed
+    // at the same level.
     private final Range<Integer> range;
     private final List<String> existingSuppressions;
     private final String suffix;
@@ -50,7 +54,14 @@ final class LazySuppressionReplacement extends Replacement {
         this.sourceCode = sourceCode;
         this.suppressWarnings = suppressWarnings;
 
-        // Calculate range immediately to avoid tree representation changes
+        // There is an additional issue that by the time error-prone comes around to apply the replacements, the
+        // compiler seems to change the representation of the tree for another phase - `App.Builder` becomes
+        // `App$Builder` etc and the start position for the expression changes to be after `App` rather than at the
+        // start of `App`. If we calculate the replacement range too late, we insert our @SuppressWarnings at the
+        // wrong location, and the indentation is miscalculated. But we can't calculate the replacement string
+        // straight away, as we might not have all the new suppressions added yet. So we have to immediately
+        // calculate the replacement range and indentation/suffix, but hold off building the final replacement
+        // string until we have all the new suppressions.
         this.range = calculateRange(endPositions, suppressWarnings, declaration);
 
         this.suffix = suppressWarnings
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
@@ -27,42 +27,37 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-final class SuppressingReplacement extends Replacement {
-    // We really do need to be this lazy for generating the Replacements, as error-prone immediately converts
-    // the Fix to a Replacement when a Description is given to it, and we need to defer the computation of the
-    // Replacement until a number of Descriptions have been produced, to handle multiple errors being suppressed
-    // at the same level.
-
+/**
+ * A Replacement that handles @SuppressWarnings modifications with line-removing capabilities.
+ * When the final replacement text is empty (no suppressions), it will remove preceding
+ * whitespace up to and including the newline.
+ */
+final class LazySuppressionReplacement extends Replacement {
     private final Range<Integer> range;
     private final List<String> existingSuppressions;
     private final String suffix;
-    private final Set<String> newSuppressions;
+    private final Set<String> desiredSuppressions;
+    private final Optional<CharSequence> sourceCode;
+    private final Optional<? extends AnnotationTree> suppressWarnings;
 
-    SuppressingReplacement(
+    LazySuppressionReplacement(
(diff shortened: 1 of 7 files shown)


suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
```diff
@@ -27,42 +27,37 @@
...
+        this.suppressWarnings = suppressWarnings;
 
-        // There is an additional issue that by the time error-prone comes around to apply the replacements, the
-        // compiler seems to change the representation of the tree for another phase - `App.Builder` becomes
-        // `App$Builder` etc and the start position for the expression changes to be after `App` rather than at the
-        // start of `App`. If we calculate the replacement range too late, we insert our @SuppressWarnings at the
-        // wrong location, and the indentation is miscalculated. But we can't calculate the replacement string
-        // straight away, as we might not have all the new suppressions added yet. So we have to immediately
-        // calculate the replacement range and indentation/suffix, but hold off building the final replacement
-        // string until we have all the new suppressions.
```
CRogers, line 58: Can we keep these more detailed comments?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
```diff
@@ -27,42 +27,37 @@
...
+    private final Optional<CharSequence> sourceCode;
+    private final Optional<? extends AnnotationTree> suppressWarnings;
 
-    SuppressingReplacement(
+    LazySuppressionReplacement(
             EndPosTable endPositions,
-            Set<String> newSuppressions,
+            Set<String> desiredSuppressions,
             Optional<CharSequence> sourceCode,
             Optional<? extends AnnotationTree> suppressWarnings,
```
CRogers, line 51: I know this is the existing code, but it would probably be better to take a single `Optional<SuppressWarningsInfo>` where `SuppressWarningsInfo` is a record that contains both the source code and annotation. Otherwise, we have this situation with two optionals where they must always be both empty or both present, and taking a single optional gets rid of those invalid states where one is present and one is empty.

CRogers: Wait, ignore this, the source code being present or not is not related to the annotation existing or not. It just whether `visitorState.getSourceCode()` is null or not

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
```diff
@@ -71,21 +66,40 @@ final class SuppressingReplacement extends Replacement {
...
     public String replaceWith() {
+        String result = calculateReplacementText();
+        return result;
+    }
+
+    private String calculateReplacementText() {
+        if (desiredSuppressions.isEmpty()) {
+            // No suppressions left, return empty string (line removal will be handled by range())
+            return "";
+        }
```
CRogers, line 101: I think this might remove non-errorprone suppresions. Eg, someone suppresses a checkstyle issue like so:

```java
@SuppressWarnings({"SomeErrorProne", "checkstyle:RegexpSinglelineJava"})
public void method() {
    System.out.println("hello");
}
```

...and we get zero errorprone matches on it (after say there is an autofix added for `SomeErrorProne`), we'd try to remove all the suppressions because our `desiredSuppressions` are empty:

```java
public void method() {
    // This is now a checkstyle failure.
    System.out.println("hello");
}
```

I think we need to only remove non-errorprone suppressions. This means we need to get all the possible `BugChecker` names (including the non-canonical ones). There are two types of `BugChecker`s:
1. Custom ones like we write. I think these can all be service loaded eg like [we do here in gradle-guide](https://github.com/palantir/gradle-guide/blob/a626b8e1cfeddcd1c34decde5712e4955adae2a0/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/utils/AllErrorprones.java#L26).
2. Built-in checks. We should be able to get these from the [`BuiltInCheckerSuppliers`](https://github.com/google/error-prone/blob/49b3262ce5b08bfbd560a96a7608099074aff844/core/src/main/java/com/google/errorprone/scanner/BuiltInCheckerSuppliers.java#L646) class

CRogers: Definitely need a test for this too.

kelvinou01: Thanks, this was a really important catch!

We have to be careful to initialize `LazySuppressingFix` with any existing non-rollout suppressions, rather than with nothing. 

I've changed `VisitorStateModifications` to the correct behavior, and added a test which fails before the change and passes after.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingReplacement.java
kelvinou01: Yep apologies!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -114,12 +116,22 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
-                        Optional.ofNullable(visitorState.getSourceCode()), suppressWarnings, firstSuppressibleParent));
-
-        suppressingFix.addSuppression(description.checkName);
+        LazySuppressionFix suppressingFix = FIXES.computeIfAbsent(firstSuppressibleParent, _ignored -> {
+            // Initialize every fix with all remove-rollout: suppressions removed
+            Stream<String> existingSuppressions = suppressWarnings
+                    .map(AnnotationUtils::annotationStringValues)
+                    .orElseGet(Stream::of);
+            Set<String> existingNonRolloutSuppressions = existingSuppressions
+                    .filter(sup -> !sup.startsWith(CommonConstants.AUTOMATICALLY_ADDED_PREFIX))
```
CRogers, line 125: So the trick to avoid removing non-errorprone suppressions is to just handle `for-rollout` ones? I guess this is fine as we haven't yet added the feature that works on human authored suppressions - we can handle that functionality when we get to it in a later PR.

CRogers (approved):

PR #200, opened 2025-10-16
Add the ability to ignore suppressions in `SuppressibleTreePathScanner`

kelvinou01:
## Before this PR

We are working on `-PerrorProneRemoveUnused`, a mode which removes all unused suppressions. This mode requires us to turn on the ignore suppressions option in error-prone, or else we can't descend into suppressed trees to chec. 

Error-prone's `SuppressibleTreePathScanner` doesn't respect this flag. This is not expected behavior, as [`ErrorProneScanner` respects the flag](https://github.com/google/error-prone/blob/ac2eaec7616878ad3f2ccfb8d964892146257b39/check_api/src/main/java/com/google/errorprone/scanner/ErrorProneScanner.java#L503). We've opened https://github.com/google/error-prone/pull/5255, but I think it will take some time to be merged. In the mean time, let's add the option to insert some bytecode so the `SuppressibleTreePathScanner` respects this flag when we need it to.

## After this PR

==COMMIT_MSG==
Add the ability to ignore suppressions in `SuppressibleTreePathScanner`
==COMMIT_MSG==

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 import one.util.streamex.EntryStream;
 
 /**
@@ -57,8 +56,7 @@ public interface CommonOptions {
             public Map<String, String> extraErrorProneCheckOptions() {
                 return EntryStream.of(CommonOptions.this.extraErrorProneCheckOptions())
                         .append(other.extraErrorProneCheckOptions())
-                        .collect(Collectors.toMap(
-                                Map.Entry::getKey, Map.Entry::getValue, (val1, val2) -> val1 + "," + val2));
+                        .toMap();
             }
 
             @Override
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import one.util.streamex.EntryStream;
 
 /**
@@ -41,6 +42,10 @@ public interface CommonOptions {
         return RemoveRolloutCheck.DISABLE;
     }
 
+    default boolean ignoreSuppressionAnnotations() {
+        return false;
+    }
+
     default CommonOptions naivelyCombinedWith(CommonOptions other) {
         return new CommonOptions() {
             @Override
@@ -52,12 +57,18 @@ public interface CommonOptions {
             public Map<String, String> extraErrorProneCheckOptions() {
                 return EntryStream.of(CommonOptions.this.extraErrorProneCheckOptions())
                         .append(other.extraErrorProneCheckOptions())
-                        .toMap();
+                        .collect(Collectors.toMap(
+                                Map.Entry::getKey, Map.Entry::getValue, (val1, val2) -> val1 + "," + val2));
             }
 
             @Override
             public RemoveRolloutCheck removeRolloutCheck() {
                 return CommonOptions.this.removeRolloutCheck().or(other.removeRolloutCheck());
+            }
+
+            @Override
+            public boolean ignoreSuppressionAnnotations() {
+                return CommonOptions.this.ignoreSuppressionAnnotations() || other.ignoreSuppressionAnnotations();
             }
         };
     }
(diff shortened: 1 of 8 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -52,13 +57,19 @@ public PatchChecksOption patchChecks() {
             public Map<String, String> extraErrorProneCheckOptions() {
                 return EntryStream.of(CommonOptions.this.extraErrorProneCheckOptions())
                         .append(other.extraErrorProneCheckOptions())
-                        .toMap();
+                        .collect(Collectors.toMap(
+                                Map.Entry::getKey, Map.Entry::getValue, (val1, val2) -> val1 + "," + val2));
```
CRogers: `EntryStream` has a slightly shorter `.toMap(mergeFunction)` method.

CRogers: So this is letting us natively has the same option multiple times, which then builds them up as a list (aka comma separated string argument)?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -47,19 +50,16 @@ static ModifyCheckApiOption noEffect() {
     }
 
     /**
-     * The {@code error_prone_check_api} jar must be modified to allow {@code `for-rollout:`} suppressions to work.
+     * Modify these files in the error-prone API.
      */
-    static ModifyCheckApiOption mustModify() {
-        return new MustModify(false);
+    static MustModify mustModify(ModifiedFile... modifiedFile) {
+        return new MustModify(Arrays.stream(modifiedFile).collect(Collectors.toSet()));
```
CRogers: `Set.of(modifiedFile)` should do the trick.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
```diff
@@ -69,9 +69,11 @@ enum NoEffect implements ModifyCheckApiOption {
         INSTANCE
     }
 
-    record MustModify(boolean modifyVisitorState) implements ModifyCheckApiOption, CombinedValue {
+    record MustModify(Set<ModifiedFile> modifiedFiles) implements ModifyCheckApiOption, CombinedValue {
         public MustModify combine(MustModify other) {
-            return new MustModify(modifyVisitorState || other.modifyVisitorState);
+            Set<ModifiedFile> union = Stream.concat(modifiedFiles.stream(), other.modifiedFiles.stream())
+                    .collect(Collectors.toSet());
```
CRogers: Streamex is a bit nicer here: `StreamEx.of(modifiedFiles).append(other.modifiedFiles).toSet()`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/SuppressibleTreePathScannerClassVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+
+            // If condition is true, return false immediately
+            Label continueLabel = new Label();
+            mv.visitJumpInsn(Opcodes.IFEQ, continueLabel);
+            mv.visitInsn(Opcodes.ICONST_0); // push false
+            mv.visitInsn(Opcodes.IRETURN);
+            mv.visitLabel(continueLabel);
+            // Add a frame here to fix verification
+            mv.visitFrame(Opcodes.F_SAME, 0, null, 0, null);
+        }
```
CRogers, line 84: ~~warms my heart~~ uh, I mean a necessary evil

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -79,18 +81,24 @@ private Optional<UnaryOperator<ClassVisitor>> classVisitorFor(String classJarPat
...
         // it intercepts all errors and changes them. So when we're just running normal compilation, we want
         // to avoid running our modifications at all, and let the errors continue on their way unchanged.
         if (classJarPath.equals("com/google/errorprone/VisitorState.class")
-                && getParameters().getModifyVisitorState().get()) {
+                && getParameters().getFilesToModify().get().contains(ModifiedFile.VISITOR_STATE)) {
             return Optional.of(VisitorStateClassVisitor::new);
         }
 
+        if (classJarPath.equals("com/google/errorprone/bugpatterns/BugChecker$SuppressibleTreePathScanner.class")
+                && getParameters().getFilesToModify().get().contains(ModifiedFile.SUPPRESSIBLE_TREE_PATH_SCANNER)) {
```
CRogers: I'd consider extracting `getParameters().getFilesToModify().get()` to a variable to make these three checks a bit clearer.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
CRogers: Why do we actually need this change?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/ModifyCheckApiOption.java
kelvinou01: haha i've always wondered why `toSet()` isn't in std, since theres `toList()`. Thanks for the tip!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
kelvinou01: Oops I think this was required to handle some extra options in the next PR. Can't remember why exactly it was needed, but let's remove it from this PR.

CRogers (approved):

Issue #173, opened 2025-09-30
Removing unnecessary `@SuppressWarnings` at the same time as suppressing


kelvinou01:
### Background

An important feature is to remove unnecessary errorprone suppressions. When writing new errorprones, it’s common to make a mistake or bug which causes the errorprone to be auto-suppressed everywhere. When a fix is published, by default the suppression just stays for ever and is not removed. This is bad as it makes it look like the issue is more widespread than it really is, and makes it easy to add the issue to same block of code and not get a warning about it.

### Current state

We can currently remove unused errorprones for for-rollout suppressions like so:

```bash
./gradlew compileAllErrorProne -PerrorProneRemoveRollout
./gradlew compileAllErrorProne -PerrorProneSuppress
```

This first removes all @SuppressWarnings("for-rollout:*") and then re-suppresses them.

However, it runs compilation twice, which is very inefficient. In some repos, compilation can take 10+ mins, so doing it twice can easily break the default excavator timeout (which we’ve temporarily raised to 25 mins).

### How to improve?

Ideally, we’d be able to do all the suppressible-error-prone steps in one compilation, eg:

```bash
./gradlew compileAllErrorProne \
    -PerrorProneRemoveUnnecessarySuppressions \
    -PerrorProneApply \
    -PerrorProneSuppress
```

This makes it efficient to remove unnecessary suppressions again.

We can also remove unnecessary human-authored suppressions rather than just automatic for-rollout ones like today.

### How to do this?

Most of the infrastructure to do this already exists. We already have a way of [intercepting errorprone Descriptions](https://github.com/palantir/suppressible-error-prone/blob/0f3c7b59f4faaa3030873c897fddf5304087a7e5/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java#L50) and [modifying/adding the SuppressWarnings](https://github.com/palantir/suppressible-error-prone/blob/d6819c781c3667b5d0dc0da6f0d048ea01516401/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressingFix.java#L32). We should be able to just add a Set<String> encounteredErrors to SuppressingFix and record all the errors/warnings that appear on the SuppressWarnings (we will have to think carefully about all the places a SuppressWarnings could be - it could be on the whole class rather than just the most specific possible place). When it comes to actually rendering the fix, we can remove any unnecessary ones in the annotations.

[...]

PR #202, opened 2025-10-17
Add `removeUnused` mode 

kelvinou01:
## Before this PR

See https://github.com/palantir/suppressible-error-prone/issues/173 for background

FLUP to #192, #200, #203 

Features
- `removeUnused` mode removes all unused error-prone suppressions (both human-authored and for-rollout suppressions) in all compilation units
- Only removes error-prone suppressions, and ignores unknown ones like checkstyles or `SomeRandomSuppression`
- Works seamlessly with apply and suppress mode, allowing you to remove unused + apply + suppress in one compilation

## After this PR

==COMMIT_MSG==
Add `removeUnused` mode 
==COMMIT_MSG==

## Possible downsides?

- Adds yet another layer of complexity onto `VisitorStateModifications`, e.g. the `SuppressibleErrorProne` check used to prevent infinite recursion when calling `state.reportMatch`
- Maybe performance regressions?

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -1644,6 +1644,43 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         '''.stripIndent(true)
     }
 
+    def 'errorProneRemoveUnused does not apply fixes'() {
+        given:
+        // language=Java
+        def initialSource = '''
+            package app;
+
+            public final class App {
+                class Inner {
+                    class InnerInner {}
+                }
+            }
+        '''.stripIndent(true)
+
+        writeJavaSourceFileToSourceSets initialSource
+
+        when: 'errorProneRemoveUnused is run by itself'
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused')
+
+        then: 'No checks are applied'
+        javaSourceIsSyntacticallyEqualTo initialSource
+
+        when: 'ClassCanBeStatic is applied alongside errorProneRemoveUnused'
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused', '-PerrorProneApply=ClassCanBeStatic')
+
+        then: 'ClassCanBeStatic is applied'
+        // language=Java
+        javaSourceIsSyntacticallyEqualTo '''
+            package app;
+
+            public final class App {
+                static class Inner {
+                    static class InnerInner {}
+                }
+            }
+        '''.stripIndent(true)
+    }
+
     def 'errorProneRemoveUnused only keeps the closest suppression to a violation'() {
         // language=Java
         writeJavaSourceFileToSourceSets '''
@@ -1668,7 +1705,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         '''.stripIndent(true)
 
         when:
-        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused')
+        def result = runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused')
 
         then:
 
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -1641,7 +1641,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     new int[3].toString();
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused only keeps the closest suppression to a violation'() {
@@ -1689,7 +1689,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     }
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused handles multiple suppressions on different tree types gracefully'() {
@@ -1751,7 +1751,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     }
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused removes entire SuppressWarnings annotation when all suppressions are unused'() {
@@ -1780,7 +1780,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                 System.out.println("No violations here");
             }
         }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused and errorProneSuppress uses existing suppressions if possible'() {
@@ -1838,7 +1838,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     }
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused and errorProneApply applies fixes on previously suppressed elements'() {
@@ -1895,7 +1895,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     }
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'errorProneRemoveUnused + errorProneApply + errorProneSuppress applies fixes and suppressions on previously suppressed elements'() {
@@ -1956,7 +1956,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     }
                 }
             }
-    '''.stripIndent(true)
+        '''.stripIndent(true)
     }
 
     def 'error-prone dependencies have versions bound together by a virtual platform'() {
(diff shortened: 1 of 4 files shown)


kelvinou01:
FLUPS: https://github.com/palantir/suppressible-error-prone/pull/203#discussion_r2444789547

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1611,6 +1610,355 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+
+            @SuppressWarnings({"ArrayToString", "InlineTrivialConstant", "NotAnErrorProne", "checkstyle:Bla"})
+            public final class App {
+                private static final String EMPTY_STRING = "";
+ 
+                public static void main(String[] args) {
+                    new int[3].toString();
+                }
+            }
+    '''.stripIndent(true)
```
CRogers: nit: indent

CRogers: \+ other instances in other tests

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1611,6 +1610,355 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+        // language=Java
+        javaSourceIsSyntacticallyEqualTo '''
+            package app;
+
+            public final class App {
+                @SuppressWarnings("InlineTrivialConstant")
+                private static final String EMPTY_STRING = "";
+
+                // Doesn't move an already existing suppression, even if it could be closer to the violation
+                @SuppressWarnings("ArrayEquals")
```
CRogers, line 1776: Interesting - shouldn't it though? I guess `removeUnused` is not correct here, but I would have thought the natural result of ascending the tree path for each Description to find the first suppressible tree to add the suppression to would move it to be closer.

CRogers: I guess this would have the downside of changing human authored suppressions? Perhaps someone suppresses a check on a whole class, it would be strange for us to force that to be different. I guess this is behaviour we want, for human suppressions at least? You could argue we want to make `for-rollout:` ones as specific as possible?

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1611,6 +1610,355 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
...
+        // language=Java
+        writeJavaSourceFileToSourceSets '''
+        package app;
+
+        @SuppressWarnings({"UnusedVariable", "ArrayToString"})
+        public final class App {
+            public static void main(String[] args) {
+                System.out.println("No violations here");
+            }
+        }
```
CRogers, line 1804: Indent the code for readability.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -21,29 +21,39 @@
...
 import java.util.stream.Stream;
 
 // CHECKSTYLE:ON
 
+@SuppressWarnings("RestrictedApi")
 public final class VisitorStateModifications {
     private static final Logger log = Logger.getLogger(VisitorStateModifications.class.getName());
-
     private static final ReportedFixCache FIXES = new ReportedFixCache();
+    private static final Set<CompilationUnitTree> SEEN = new HashSet<>();
```
CRogers: Let's give `SEEN` a slightly better name - maybe `ALREADY_REMOVED_EXISTING_SUPPRESSIONS`?

CRogers: Actually, I think we should make a new type to remove some code from `VisitorStateModifications`. It can hold and manage this `Set`, then have the `removeAllSuppressionsOnErrorpronesInCompilationUnit` method in there too.

CRogers: Similar to `ReportedFixCache` (will need to take it as a field)

CRogers: This `HashSet` will make strong reference on each `CompilationUnitTree` which will cause a memory leak. We should make this a weak set instead. There is no `WeakSet` type but you can use `Collections.newSetFromMap(new WeakHashMap<>())` to make one.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -74,6 +84,25 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+            return Description.NO_MATCH;
+        }
+
+        if (!modes.contains("Suppress")) {
+            log.warning("No autofix was found for " + description.checkName + " at position "
+                    + description.position.getStartPosition() + " in "
+                    + visitorState.getPath().getCompilationUnit().getSourceFile() + "."
+                    + " -PerrorProneSuppress was not passed either. "
+                    + " SuppressibleErrorProne will not be able to add a suppression for this error.");
+            return Description.NO_MATCH;
```
CRogers, line 103: I'm probably missing something, but wouldn't it be fine to just have this remain an error/warning etc when the suppression mode is not enabled?

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
kelvinou01: > Interesting - shouldn't it though?

Suppressions done too high on the tree hierarchy — "overbroad" suppressions — are always human-made. The rationale behind not moving them was the machine shouldn't interfere when a human deliberately placed the suppression. 

e.g. we might roll something some check X which is not applicable for an entire file in a consumer repo. Devs remove all for-rollout:X suppressions scattered in said file and put a suppression on the class declaration. i think we should respect that; it would annoy the devs if we undid their hard work the next time we came along and `removeUnused`

```java
SuppressWarnings("X") // not applicable because blablabla
public class SomethingUnapplicableForX { ... }
```

kelvinou01: > You could argue we want to make for-rollout: ones as specific as possible?

We could, as a special case for removeUnused + suppress mode. It'd add a bit of complexity to [the code where we attach fixes](https://github.com/palantir/suppressible-error-prone/pull/202/files#diff-135685150f19effc6d9cd8304d7a1e84eae945f9ac03e2d5747a53a1f7f8b657R87-R95), so I'm hesitant to add it to this PR. We could do this as a FLUP if needed.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
kelvinou01: If we return `description`, an autofix might circumnavigate this if-statement, and be applied when we don't have ApplyMode on. 

```java
        boolean shouldPreferDefaultSuggestedFixesForThisCheck = patchChecks.contains(description.checkName);
        boolean checkHasSuggestedFixes = !description.fixes.isEmpty();

        if (shouldPreferDefaultSuggestedFixesForThisCheck && checkHasSuggestedFixes) {
            return description;
        }
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
kelvinou01: Yep, I've created a `SuppressionRemover` which encapsulates a weak set and the "report on everything" logic.

kelvinou01: Good catch, I didn't think of making a weak set from a weak map like this!

gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
CRogers: as discussed internally:

> Suppressions done too high on the tree hierarchy — “overbroad” suppressions — are always human-made.

What about the case where the code changes around an automated suppression, eg starting with this after suppressing:
```java
@SuppressWarnings("for-rollout:ArrayToString")
void blah() {
    new int[3].toString();
}
```
but someone comes along and adds a variable:
```java
@SuppressWarnings("for-rollout:ArrayToString")
void blah() {
    String string = new int[3].toString();
}
```
ideally it should now be:
```java
void blah() {
    @SuppressWarnings("for-rollout:ArrayToString")
    String string = new int[3].toString();
}
```

CRogers: Agreed to follow up with this behaviour for `for-rollout` in a later PR.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
CRogers: Ok, if I'm understanding this properly (I tried to last night by my brain was pretty frazzled) the issue is we might apply autofixes when in `removeUnused` mode but without `apply` mode activated?

I changed this to just `return description` and the test "errorProneRemoveUnused only keeps the closest suppression to a violation" failed, which is good and helped me a bit, but I think it only did so "accidentally" and it's not super clear (it changes non-static classes to be static). I think it would be very helpful for me/future people to have an explicit test for this case that sets up the failure in a way that's super obvious from reading the test.

CRogers: I think it also needs a comment in the source.

kelvinou01: > Ok, if I'm understanding this properly (I tried to last night by my brain was pretty frazzled) the issue is we might apply autofixes when in removeUnused mode but without apply mode activated?

Pretty much. 

The silver lining here is after we get rid of the concept of PreferPatchChecks, -XoptPatchChecks will regain it's original meaning, and the problem of unintended autofixes won't happen anymore.

CRogers:
:+1:

Issue #180, opened 2025-10-02
Have each suppressible-error-prone `Mode` self-contain it's semantics


kelvinou01:
### Background

We recently introduced the concept of `Mode`s. When certain modes are run together, they require special handling through `ModeInterference`s. This refactor significantly reduced the complexity of `SuppressibleErrorPronePlugin`.

However, one limitation of the current model is that it isn't immediately obvious, when reading just the mode and interference classes, what it means for the behavior of error-prone. For example, consider what happens when you run `-PerrorProneApply` and `-PerrorProneSuppress` together:

```java
public final class ApplyMode implements Mode {
    @Override
    public CommonOptions configureAndReturnCommonOptions(ModeOptionContext context) {
        return new CommonOptions() {
            @Override
            public PatchChecksOption patchChecks() {
                return PatchChecksOption.someChecks(() -> checksToApplySuggestedPatchesFor(context));
            }
        };
    }

    private static Set<String> checksToApplySuggestedPatchesFor(ModeOptionContext context) {
        boolean hasSpecificPatchChecks =
                context.flagValue().isPresent() && !context.flagValue().get().isBlank();

        if (hasSpecificPatchChecks) {
            return Splitter.on(',')
                    .omitEmptyStrings()
                    .splitToStream(context.flagValue().get())
                    .map(String::trim)
                    .collect(Collectors.toSet());
        }

        return context.extension().patchChecksForCompilation(context.javaCompile());
    }
}
```

```java
public final class SuppressingAndApplyingInterference extends SpecificModeInterference {
    @Override
    protected Set<ModeName> interferingModes() {
        return Set.of(ModeName.SUPPRESS, ModeName.APPLY);
    }

    @Override
    public CommonOptions interfere(Map<ModeName, CommonOptions> modeCommonOptions) {
        CommonOptions suppress = modeCommonOptions.get(ModeName.SUPPRESS);
        CommonOptions apply = modeCommonOptions.get(ModeName.APPLY);

        // If we're applying suggested patches at the same time as suppressing, we still need to tell
        // errorprone to patch all checks, so we can make suggested fixes for suppressions in any check.
        // However, inside our changes to errorprone, we need to get the list of checks that we're going
        // to use the default suggested fixes for, so we can work out which ones to use the suggested
        // fixes for and which to suppress. So we add the PreferPatchChecks argument here, which we can
        // use inside error-prone/the compiler.

        return suppress.naivelyCombinedWith(apply)
[...]

Issue #178, opened 2025-10-02
Make it easy for libraries to ship their own BugCheckers


kelvinou01:
### Background

At Palantir, we're increasingly reliant on error-prone to automatically mitigate common correctness/safety pitfalls — it is an effective force multiplier for small infra teams. Two patterns that have emerged are
1. Libraries shipping their own BugCheckers to enforce correct use of their APIs (see this [internal library](https://pl.ntr/2xf))
2. Dedicated code quality repos which enforce correct use of APIs of other libraries (see [gradle-guide-error-prone](https://github.com/palantir/gradle-guide/tree/develop/gradle-guide-error-prone), [baseline-error-prone](https://github.com/palantir/gradle-baseline/tree/develop/baseline-error-prone)). 

Currently, there is considerable friction for libraries to ship their own error-prones. A library has to create a new error-prone project, create a Gradle plugin which applies the [SuppressibleErrorPronePlugin](https://github.com/palantir/suppressible-error-prone/blob/develop/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java), and add the error-prone jar as a dependency. Consumers of the library are trusted that alongside using the library, they will also apply the <libraryname>-error-prone plugin to enforce good behavior. 

### Desired behavior

Library authors shouldn't have to ship a gradle plugin. As long as the library consumer has suppressible-error-prone applied, the library's BugCheckers should be automatically ran during compilation.

kelvinou01:
As discussed in person, we still want people to publish a separate jar for error-prones. Because adding too much stuff to annotation classpath has caused many problems in the past, e.g. using two different versions of the same dependencies (?)

Another reason is that using `javaVersions { javaCompiler }` + `jvmModuleArgs` (required for authoring checks) means we don't have `--release`. We want to limit any potential runtime errors to the error-prone module, so they happen at error-prone runtime (i.e. compile time)

One potential solution is

- A library ships both the my-library jar and an my-library-error-prone jar
- Consumers which apply SuppressibleErrorPronePlugin will also bring in ApplyLibraryErrorPrones plugin
- The ApplyLibraryErrorProne plugin will see "oh, you have my-library. now do variant reselection to find the dependency with `shipsErrorProne=true`". This will get the my-library-error-prone jar. Then add this dep to the `errorprone` configuration

kelvinou01:
[...]

PR #203, opened 2025-10-17
Add all the supporting infrastructure for `removeUnused` mode

kelvinou01:
## Before this PR

See https://github.com/palantir/suppressible-error-prone/issues/173 for background

FLUP to #192 and #200 

We are working on `-PerrorProneRemoveUnused`, a mode which removes all unused suppressions. This is the penultimate PR which brings in all the infrastructure required to support removeUnused mode:

- New modes and interferences (`removeUnused` leads to UnsupportedOperationException for the time being)
- `ReportedFixCache` to encapsulate creating and reporting new suppression fixes. Note that this uses `state.reportMatch`, c.f. the [old mechanism](https://github.com/palantir/suppressible-error-prone/blob/develop/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java#L141) of returning a populated `Description` from `interceptDescription`
- A new `SuppressibleErrorProne` bugchecker, used to prevent infinite recursion from `ReportedFixCache` calling `state.reportMatch`
- Various utils 

The actual implementation for removeUnused will be brought in the next PR for ease of review.

## After this PR

==COMMIT_MSG==
Add the `removeUnused` mode and it's interferences
==COMMIT_MSG==


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/common/CommonOptions.java
```diff
@@ -56,7 +57,8 @@ public PatchChecksOption patchChecks() {
             public Map<String, String> extraErrorProneCheckOptions() {
                 return EntryStream.of(CommonOptions.this.extraErrorProneCheckOptions())
                         .append(other.extraErrorProneCheckOptions())
-                        .toMap();
+                        .collect(Collectors.toMap(
```
kelvinou01, line 60: We have to specify how we collect duplicate keys here, or else we get an error

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveUnusedMode.java
```diff
@@ -0,0 +1,54 @@
...
+        return new CommonOptions() {
+            @Override
+            public PatchChecksOption patchChecks() {
+                return PatchChecksOption.allChecks();
+            }
+
+            @Override
+            public Map<String, String> extraErrorProneCheckOptions() {
+                return Map.of("SuppressibleErrorProne:Mode", "RemoveUnused");
+            }
```
kelvinou01, line 46: This will be used in VisitorStateModifications to branch to the removeUnused procedure.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/ApplyMode.java
```diff
@@ -31,6 +32,11 @@ public CommonOptions configureAndReturnCommonOptions(ModeOptionContext context)
             public PatchChecksOption patchChecks() {
                 return PatchChecksOption.someChecks(() -> checksToApplySuggestedPatchesFor(context));
             }
+
+            @Override
+            public Map<String, String> extraErrorProneCheckOptions() {
+                return Map.of("SuppressibleErrorProne:Mode", "Apply");
+            }
```
kelvinou01, line 39: To maintain parity with RemoveUnusedMode. Also, potentially a way to move mode interference logic into VisitorStateModifications in the future (see https://github.com/palantir/suppressible-error-prone/issues/180)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
```diff
@@ -49,7 +49,9 @@ public Map<String, String> extraErrorProneCheckOptions() {
 
                 return Map.of(
                         "SuppressibleErrorProne:RemoveRolloutSuppressions",
-                        context.flagValue().orElse(ALL_CHECKS));
+                        context.flagValue().orElse(ALL_CHECKS),
+                        "SuppressibleErrorProne:Mode",
+                        "RemoveRollout");
```
kelvinou01: To maintain parity with RemoveUnusedMode. Also, potentially a way to move mode interference logic into VisitorStateModifications in the future (see https://github.com/palantir/suppressible-error-prone/issues/180)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/SuppressMode.java
```diff
@@ -35,6 +36,11 @@ public CommonOptions configureAndReturnCommonOptions(ModeOptionContext context)
             public PatchChecksOption patchChecks() {
                 return PatchChecksOption.allChecks();
             }
+
+            @Override
+            public Map<String, String> extraErrorProneCheckOptions() {
+                return Map.of("SuppressibleErrorProne:Mode", "Suppress");
+            }
```
kelvinou01, line 43: To maintain parity with RemoveUnusedMode. Also, potentially a way to move mode interference logic into VisitorStateModifications in the future (see https://github.com/palantir/suppressible-error-prone/issues/180)

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemoveUnusedAndSuppressInterference.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    protected Set<ModeName> interferingModes() {
+        return Set.of(ModeName.REMOVE_UNUSED, ModeName.SUPPRESS);
+    }
+
+    @Override
+    public CommonOptions interfere(Map<ModeName, CommonOptions> modeCommonOptions) {
+        CommonOptions removeUnused = modeCommonOptions.get(ModeName.REMOVE_UNUSED);
+        CommonOptions suppress = modeCommonOptions.get(ModeName.SUPPRESS);
+        return removeUnused.naivelyCombinedWith(suppress);
```
CRogers: Is this interference doing anything? It's just naively combining?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemoveUnusedAndApplyingInterference.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.palantir.gradle.suppressibleerrorprone.modes.interferences;
+
+import com.palantir.gradle.suppressibleerrorprone.modes.common.CommonOptions;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.ModeName;
+import com.palantir.gradle.suppressibleerrorprone.modes.common.SpecificModeInterference;
+import java.util.Map;
+import java.util.Set;
+
+public final class RemoveUnusedAndApplyingInterference extends SpecificModeInterference {
```
CRogers: Isn't this identical to the suppressing + applying interference? What happens when we have suppressing + applying + removeUnused?

I'm honestly not convinced that this interference concept was necessary (mea culpa) - I think we may have been able to always use `-XepPatchChecks:` to enable patching on every bugchecker and always send a list of checks that need patching to `VisitorStateModifications`, which would remove fixes if it is not to be patched (when suppressing/applying/etc).

If we still need to keep interferences, I think we could use the `ModeInterference` interface directly, return all of suppress/apply/removeUnused as interfering if they have been enabled. The combine all of them together in `interfere` (and only have one interference between all three).

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -0,0 +1,43 @@
...
+                ServiceLoader.load(BugChecker.class).stream().map(ServiceLoader.Provider::get);
+        Stream<String> pluginCheckNames =
+                StreamEx.of(pluginChecks).flatMap(bugchecker -> bugchecker.allNames().stream());
+
+        Stream<BugCheckerInfo> builtInChecks = BuiltInCheckerSuppliers.allChecks().getAllChecks().values().stream();
+        Stream<String> builtInCheckNames =
+                StreamEx.of(builtInChecks).flatMap(bugchecker -> bugchecker.allNames().stream());
+
+        return Stream.concat(pluginCheckNames, builtInCheckNames).collect(Collectors.toSet());
+    }
```
CRogers: If we're going to be repeatedly using this, we definitely want to memoize it by backing it with a static `Suppliers.memoize`.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
```diff
@@ -0,0 +1,101 @@
...
+
+final class ReportedFixCache {
+    // Weak map so that we don't leak memory by keeping hold of references to the source element tree keys and our
+    // mutable fixes values around forever, once error-prone has finished with the source element tree used as a key
+    // here (once the file has been visited by all the error-prone checks), our SuppressingFixes can be safely
+    // garbage collected.
+    private final WeakHashMap<Tree, LazySuppressionFix> cache = new WeakHashMap<>();
+
+    public static final Predicate<String> REMOVE_EVERYTHING = bc -> false;
+    public static final Predicate<String> KEEP_EVERYTHING = bc -> true;
```
CRogers, line 42: Are these used?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
```diff
@@ -0,0 +1,101 @@
...
+    @SuppressWarnings("PreferSafeLoggableExceptions") // It doesn't matter in our internal codebases
+    public LazySuppressionFix reportNew(
+            TreePath declaration, VisitorState visitorState, Predicate<String> filterExisting) {
+        if (cache.containsKey(declaration.getLeaf())) {
+            throw new IllegalArgumentException("A fix on this declaration already exists");
+        }
+        return cache.put(declaration.getLeaf(), createAndReportFix(declaration, visitorState, filterExisting));
+    }
+
+    @SuppressWarnings("RestrictedApi")
```
CRogers: I'd put this suppression right on the `Description.builder` var declaration to make it more obvious what it applies to.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
CRogers: Or will this be used in the next PR?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressibleErrorProne.java
```diff
@@ -0,0 +1,32 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/suppressible-error-prone",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "A change made by suppressible-error-prone",
+        // Make it unsuppressible so that it can actually remove itself
+        suppressionAnnotations = {})
+public class SuppressibleErrorProne extends BugChecker {}
```
CRogers, line 37: What is this used for?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
kelvinou01: These are for convenience and to document what the predicate does. `KEEP_EVERYTHING` is used in the next PR.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressibleErrorProne.java
kelvinou01: We have to pass in a valid BugChecker name in `ReportedFixCache`. I don't think using e.g. the name of the suppressions being removed is correct here, as a fix may remove multiple suppressions or no suppressions, so I created this default BugChecker. 

I'll document this under the class.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
kelvinou01: Good idea. The computation here doesn't involve the VisitorState, so I used a simple memoized getter instead.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemoveUnusedAndApplyingInterference.java
kelvinou01: Now that I think about it, if we are reporting rollout suppressions with `SuppressibleErrorProne` bugchecker, I think we can get rid of `PreferPatchChecks` entirely. `-XepPatchChecks` will retain it's original semantics of actually applying the autofixes.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemoveUnusedAndSuppressInterference.java
kelvinou01: Thanks for the catch, I'll remove this

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/RemoveUnusedAndApplyingInterference.java
kelvinou01: As discussed privately, let's consolidate the duplicate logic to a `ApplyingAndSuppressingOrRemoveUnusedInterference`

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/ApplyingAndSuppressingOrRemoveUnusedInterference.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final Set<ModeName> APPLY_AND_REMOVE_UNUSED = Set.of(ModeName.APPLY, ModeName.REMOVE_UNUSED);
+    private static final Set<ModeName> APPLY_AND_SUPPRESS = Set.of(ModeName.APPLY, ModeName.SUPPRESS);
+
+    @Override
+    public ModeInterferenceResult interferesWith(Set<ModeName> modeNames) {
+        Optional<Set<ModeName>> maximalInterference = getMaximalInterference(modeNames);
+        return maximalInterference
+                .map(ModeInterferenceResult::interferenceBetween)
+                .orElseGet(ModeInterferenceResult::noInterference);
+    }
```
CRogers: I think this might be more readable/less writing out all the possibilities as (untested):

```java
if (!modeNames.contains(ModeName.APPLY) {
    return ModeInterferenceResult.noInterference();
}

Set<ModeName> interferingWithApply = Set.intersection(modeNames, Set.of(ModeName.SUPPRESS, ModeName.REMOVE_UNUSED));

if (interferingWithApply.isEmpty()) {
    return ModeInterferenceResult.noInterference();
}

return ModeInterferenceResult.interferenceBetween(Sets.union(Set.of(ModeName.APPLY), interferingWithApply));
```

...but think this is up to taste.

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/ApplyingAndSuppressingOrRemoveUnusedInterference.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public CommonOptions interfere(Map<ModeName, CommonOptions> modeCommonOptions) {
+        Set<ModeName> maximalInterference =
+                getMaximalInterference(modeCommonOptions.keySet()).get();
+        CommonOptions apply = modeCommonOptions.get(ModeName.APPLY);
+        CommonOptions naivelyCombined = EntryStream.of(modeCommonOptions)
+                .filterKeys(maximalInterference::contains)
+                .values()
+                .reduce(CommonOptions.empty(), CommonOptions::naivelyCombinedWith);
```
CRogers: I think this is only going to give you the mode names that you returned in `interferesWith`, so you can just combine them all then add the extra flag - no need to calculate the maximal interference again?

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/ApplyingAndSuppressingOrRemoveUnusedInterference.java
kelvinou01: Yeah, I think this reads nicer!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/interferences/ApplyingAndSuppressingOrRemoveUnusedInterference.java
kelvinou01: ahh yep, good catch!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import one.util.streamex.StreamEx;
+
+public final class AllErrorprones {
+    private static Set<String> cachedAllBugcheckerNames;
+
+    public static Set<String> allBugcheckerNames() {
+        if (cachedAllBugcheckerNames == null) {
```
CRogers: Use `Suppliers.memoized(` to set the field. It will handle concurrency/exceptions a bit better. Multiple threads should be calling this code. It's also a bit more succinct.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
```diff
@@ -0,0 +1,104 @@
...
+            TreePath declaration, VisitorState visitorState, Predicate<String> filterExisting) {
+        return cache.computeIfAbsent(
+                declaration.getLeaf(), _ignored -> createAndReportFix(declaration, visitorState, filterExisting));
+    }
+
+    /**
+     * Initialize a {@code LazySuppressionFix} on {@code declaration}, choosing which suppressions to keep with
+     * {@code filterExisting}.
+     */
+    @SuppressWarnings("PreferSafeLoggableExceptions") // It doesn't matter in our internal codebases
```
CRogers, line 61: Should probably disable it in the build.gradle for this project.

CRogers: Actually, I think the it shouldn't even suggest this unless `safe-logging` is on the classpath. Not sure why it would be.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
```diff
@@ -0,0 +1,104 @@
...
+
+        LazySuppressionFix fix = new LazySuppressionFix(
+                Optional.ofNullable(state.getSourceCode()),
+                suppressWarnings,
+                declaration.getLeaf(),
+                filteredExistingSuppressions);
+        @SuppressWarnings("RestrictedApi")
+        Description description = Description.builder(
+                        (DiagnosticPosition) declaration.getLeaf(),
+                        "SuppressibleErrorProne",
```
CRogers: To make it really obvious that it's the bug checker in this repo, could do `SuppressibleErrorProne.class.getSimpleName()`.

CRogers: Will handle refactor renames too.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -21,39 +21,30 @@
...
-    // here (once the file has been visited by all the error-prone checks), our SuppressingFixes can be safely
-    // garbage collected.
-    private static final Map<Tree, LazySuppressionFix> FIXES = new WeakHashMap<>();
+    private static final ReportedFixCache FIXES = new ReportedFixCache();
 
     @SuppressWarnings("RestrictedApi")
     public static Description interceptDescription(VisitorState visitorState, Description description) {
-        if (description == Description.NO_MATCH) {
+        // Prevent infinite recursion on reported fixes
+        if (description == Description.NO_MATCH || description.checkName.equals("SuppressibleErrorProne")) {
```
CRogers: Another `SuppressibleErrorProne.class.getSimpleName()`.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -21,39 +21,30 @@
...
-    // mutable fixes values around forever, once error-prone has finished with the source element tree used as a key
-    // here (once the file has been visited by all the error-prone checks), our SuppressingFixes can be safely
-    // garbage collected.
-    private static final Map<Tree, LazySuppressionFix> FIXES = new WeakHashMap<>();
+    private static final ReportedFixCache FIXES = new ReportedFixCache();
 
     @SuppressWarnings("RestrictedApi")
     public static Description interceptDescription(VisitorState visitorState, Description description) {
-        if (description == Description.NO_MATCH) {
+        // Prevent infinite recursion on reported fixes
```
CRogers, line 39: Nasty

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/ApplyMode.java
```diff
@@ -31,6 +32,11 @@ public CommonOptions configureAndReturnCommonOptions(ModeOptionContext context)
             public PatchChecksOption patchChecks() {
                 return PatchChecksOption.someChecks(() -> checksToApplySuggestedPatchesFor(context));
             }
+
+            @Override
+            public Map<String, String> extraErrorProneCheckOptions() {
+                return Map.of("SuppressibleErrorProne:Mode", "Apply");
```
CRogers: I would at least make `"SuppressibleErrorProne:Mode"` a constant across all these files.

It could actually be a higher level concept, ie have it's own method on `CommonOptions` or - even better - the modes framework could automatically add check option based on the modes in use, meaning this isn't even necessary.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
kelvinou01: [Seems like it's being suggested everywhere](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java). Maybe the safe logging errorprones should live in safe-logging instead after #178 is out

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
kelvinou01: Oh ok now I understand. I thought you were talking about errorprone's VisitorState::memoize so i got confused as this doesn't use the VisitorState to compute anything. Gotcha.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/ReportedFixCache.java
kelvinou01: Yep!

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/ApplyMode.java
kelvinou01: Let's make it a constant for now, and tweak the concept in [the next PR](https://github.com/palantir/suppressible-error-prone/pull/203#discussion_r2444789547)

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -103,7 +104,7 @@ public static Description interceptDescription(VisitorState visitorState, Descri
     }
 
     private static Set<String> getModes(VisitorState state) {
-        return state.errorProneOptions().getFlags().getSetOrEmpty("SuppressibleErrorProne:Mode");
+        return state.errorProneOptions().getFlags().getSetOrEmpty("SuppressibleErrorProne:Modes");
```
CRogers: Shouldn't this still be `Mode` rather than `Modes`?

kelvinou01: oops yep fixed it.

CRogers:
:+1L

CRogers:
:+1:

PR #225, opened 2025-10-31
Ensure we use the correct module args for test-error-prones

CRogers:
## Before this PR
As I was trying to fix `moduleJvmArgs` applying `--add-exports` to `compilerArgs` rather than `forkArgs`, I tested in this repo and it didn't work. It's because that bug has been hiding the fact these exports are needed.

## After this PR
Fix the exports for the testing error-prones.


felixdesouza:
👍

PR #249, opened 2025-11-14
Don't ignore all files in build directory

pkoenig10:
## Before this PR

The `com.palantir.suppressible-error-prone` plugin configures Error Prone to exclude any paths with a `build` path segment. Our [`gradle-plugin-testing`](https://github.com/palantir/gradle-plugin-testing) framework generates test projects inside the  `build` directory. So any test that attempts to apply and Error Prone check and verify that compilation is unsuccessful will fail because the source files will be ignored and the compilation will succeed.

This is also a problem with the Nebula based tests in this repo - see the small hack we do in `SuppressibleErrorPronePluginIntegrationTest` to get this to work.

## After this PR

Change the excluded paths regex to not match any path with a `build` path segment and instead only those that have `build/generated` path segments.

I don't know of any plugins that generate code in the `build` directory somewhere other than `generated`. There are some legacy plugins that generated code in `src/generated` or `generated_src`, but this PR does not affect those exclusions.

### DIFF
--- a/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
+++ b/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
@@ -239,7 +239,7 @@ public abstract class SuppressibleErrorPronePlugin implements Plugin<Project> {
         // /check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1277-L1285
 
         // language=RegExp
-        return ".*/(build|generated_.*[sS]rc|src/generated.*)/.*";
+        return ".*/(build/generated.*|src/generated.*|generated_.*[sS]rc)/.*";
     }
 
     public abstract static class PatchChecksCommandLineArgumentProvider implements CommandLineArgumentProvider {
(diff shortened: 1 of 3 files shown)


gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java
```diff
@@ -239,7 +239,7 @@ static String excludedPathsRegex() {
         // /check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1277-L1285
 
         // language=RegExp
-        return ".*/(build|generated_.*[sS]rc|src/generated.*)/.*";
+        return ".*/(build/generated.*|src/generated.*|generated_.*[sS]rc)/.*";
```
FinlayRJW, line 242: not sure if it would be safer to instead of blocking anything with `build/generated` to allow anything under `build/gradle-plugin-testing` explicitly, something like `*/(build/(?!gradle-plugin-testing).*|generated_.*[sS]rc|src/generated.*)/.*`

pkoenig10: It seems better to not reference specific plugins here. This could impact internal plugins or other things where it's less appropriate to put the name here.

FinlayRJW:
👍

pkoenig10:
FLUP to upgrade this in gradle-baseline in https://github.com/palantir/gradle-baseline/pull/3361.

PR #278, opened 2025-12-12
Fix removeUnused mode not loading plugin checks and alt names

kelvinou01:
## Before this PR

A few bugs in suppressible-error-prone:

- AllBugCheckers wasn't loading any plugin checks
- RemoveUnused mode didn't understand suppressions using alt-names, e.g. SuppressWarnings("unused"), leading it to overzealously remove suppressions

## After this PR

- AllBugCheckers loads plugin checks
- RemoveUnused mode understands alt-names

==COMMIT_MSG==
Fix removeUnused mode not loading plugin checks and alt names
==COMMIT_MSG==

## Possible downsides?

Probably next PR: RemoveUnused should not remove human-made suppressions. For example, a human-made suppression on  [SafeLoggingPropagation](https://github.com/palantir/baseline-error-prone/blob/24e41b5ab7652c72243cde7977c49c71640e0f67/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java#L66) for a method which is safe to log. RemoveUnused + Apply should not remove the suppression and add the `@Unsafe` annotation.

If needed, human-authored suppressions can be removed on an allow-list basis. But the correct and performant removal of robot-added suppressions should be the focus here.

### DIFF
--- a/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
+++ b/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
@@ -76,6 +76,8 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                 // This should guarantee that we're using the same version, both of which should be in maven local
                 //   and be the current version
                 errorprone 'com.palantir.suppressible-error-prone:test-error-prone-checks:' + project.findProperty("suppressibleErrorProneVersion")
+                errorprone "com.uber.nullaway:nullaway:0.12.12"
+                // NullAway is turned off by default, but is used in some tests
             }
             
             suppressibleErrorProne {
@@ -83,7 +85,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                     // These interfere with some tests, so disable them
                     // TODO(callumr): Rewrite the tests to use custom testing error-prones rather than built in checks
                     //                to make upgrading error-prone easier.
-                    disable('Varifier', 'ReturnValueIgnored', 'UnusedVariable', 'IdentifierName', 'UnusedMethod')
+                    disable('Varifier', 'ReturnValueIgnored', 'UnusedVariable', 'IdentifierName', 'UnusedMethod', 'NullAway')
                     ignoreUnknownCheckNames = true
                 }
             }
@@ -1596,7 +1598,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         writeJavaSourceFileToSourceSets '''
             package app;
 
-            @SuppressWarnings({"ArrayToString", "UnnecessaryFinal", "InlineTrivialConstant", "NotAnErrorProne", "checkstyle:Bla",})
+            @SuppressWarnings({"ArrayToString", "UnnecessaryFinal", "InlineTrivialConstant", "NotAnErrorProne", "ShouldBePrivate", "checkstyle:Bla",})
             public final class App {
                 private static final String EMPTY_STRING = "";
  
@@ -1607,7 +1609,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         '''.stripIndent(true)
 
         when:
-        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused=ArrayToString')
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused')
 
         then:
         // language=Java
@@ -1621,6 +1623,48 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
                 public static void main(String[] args) {
                     new int[3].toString();
                 }
+            }
+        '''.stripIndent(true)
+    }
+
+    def 'errorProneRemoveUnused understands alt-names'() {
+        // language=Java
+        writeJavaSourceFileToSourceSets '''
+            package app;
+
+            public final class App {
+                @SuppressWarnings("ShouldBePrivate")
+                void fixme(String s) {}
+
+                @SuppressWarnings("MustBePrivate")
+                void fixme(Integer s) {}
+                
+                @SuppressWarnings("ShouldBePrivate")
+                private void fixme(Float s) {}
+
+                @SuppressWarnings("MustBePrivate")
+                private void fixme(Character s) {}
+            }
+        '''.stripIndent(true)
+
+        when:
+        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused')
+
+        then:
+        // language=Java
+        javaSourceIsSyntacticallyEqualTo '''
+            package app;
+
+            public final class App {
+                @SuppressWarnings("ShouldBePrivate")
+                void fixme(String s) {}
+
+                @SuppressWarnings("MustBePrivate")
+                void fixme(Integer s) {}
+                
+                private void fixme(Float s) {}
+
+                private void fixme(Character s) {}
             }
         '''.stripIndent(true)
     }
--- a/suppressible-error-prone/build.gradle
+++ b/suppressible-error-prone/build.gradle
@@ -21,7 +21,9 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.comp',
             'jdk.compiler/com.sun.tools.javac.file',
             'jdk.compiler/com.sun.tools.javac.main',
+            'jdk.compiler/com.sun.tools.javac.model',
             'jdk.compiler/com.sun.tools.javac.parser',
+            'jdk.compiler/com.sun.tools.javac.processing',
             'jdk.compiler/com.sun.tools.javac.tree',
             'jdk.compiler/com.sun.tools.javac.util')
 }
--- a/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
+++ b/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
@@ -16,33 +16,63 @@
 
 package com.palantir.suppressibleerrorprone;
 
-import com.google.common.base.Suppliers;
 import com.google.errorprone.BugCheckerInfo;
+import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.scanner.BuiltInCheckerSuppliers;
+import com.google.errorprone.suppliers.Supplier;
+import com.sun.tools.javac.processing.JavacProcessingEnvironment;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.ServiceLoader;
+import java.util.ServiceLoader.Provider;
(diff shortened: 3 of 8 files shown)


gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy
```diff
@@ -1607,7 +1609,7 @@ class SuppressibleErrorPronePluginIntegrationTest extends ConfigurationCacheSpec
         '''.stripIndent(true)
 
         when:
-        runTasksSuccessfully('compileAllErrorProne', '-PerrorProneRemoveUnused=ArrayToString')
```
kelvinou01, line 1610: RemoveUnused doesn't take arguments — this argument was being ignored

test-error-prone-checks/src/main/java/com/palantir/gradle/suppressibleerrorprone/ShouldNotReturn.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.errorprone.bugpatterns.BugChecker.MethodTreeMatcher;
+import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.MethodTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "This check is meant only for testing suppressible-error-prone functionality")
+public final class ShouldNotReturn extends BugChecker implements MethodTreeMatcher {
```
aldexis, line 32: Did you use this anywhere?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -16,33 +16,63 @@
...
+import one.util.streamex.EntryStream;
 import one.util.streamex.StreamEx;
 
 public final class AllErrorprones {
-    private static Supplier<Set<String>> cachedAllBugcheckerNames = Suppliers.memoize(() -> {
-        Stream<BugChecker> pluginChecks =
-                ServiceLoader.load(BugChecker.class).stream().map(ServiceLoader.Provider::get);
-        Stream<String> pluginCheckNames =
-                StreamEx.of(pluginChecks).flatMap(bugchecker -> bugchecker.allNames().stream());
+    private static Supplier<Map<String, Set<String>>> canonicalToAllNames = VisitorState.memoize(state -> {
```
aldexis, line 37:
```suggestion
    private static final Supplier<Map<String, Set<String>>> canonicalToAllNames = VisitorState.memoize(state -> {
```

suppressible-error-prone/build.gradle
```diff
@@ -21,7 +21,9 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.comp',
             'jdk.compiler/com.sun.tools.javac.file',
             'jdk.compiler/com.sun.tools.javac.main',
+            'jdk.compiler/com.sun.tools.javac.model',
             'jdk.compiler/com.sun.tools.javac.parser',
+            'jdk.compiler/com.sun.tools.javac.processing',
```
aldexis, line 26: Checking out the PR locally, it seems like we might need to add these to `gradle.properties` too?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -16,33 +16,63 @@
...
+        EntryStream<String, Set<String>> builtInCheckNames =
+                StreamEx.of(builtInChecks).mapToEntry(BugCheckerInfo::canonicalName, BugCheckerInfo::allNames);
 
-        return Stream.concat(pluginCheckNames, builtInCheckNames).collect(Collectors.toSet());
+        return pluginCheckNames.append(builtInCheckNames).toMap();
     });
 
-    public static Set<String> allBugcheckerNames() {
-        return cachedAllBugcheckerNames.get();
+    private static Supplier<Set<String>> allNames =
```
aldexis, line 55:
```suggestion
    private static final Supplier<Set<String>> allNames =
```

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -16,33 +16,63 @@
...
+            VisitorState.memoize(state -> EntryStream.of(canonicalToAllNames.get(state))
+                    .flatMap(entry -> entry.getValue().stream())
+                    .collect(Collectors.toSet()));
+
+    private static Supplier<Map<String, Set<String>>> nameToPossibleCanonicalNames =
```
aldexis, line 60:
```suggestion
    private static final Supplier<Map<String, Set<String>>> nameToPossibleCanonicalNames =
```

aldexis, line 60: Is it possible to have multiple checks share the same alt name? I imagine so, though it's also pretty surprising

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/SuppressionRemover.java
```diff
@@ -40,7 +40,9 @@ public Void visitAnnotation(AnnotationTree node, Void unused) {
                     if (AnnotationUtils.isSuppressWarningsAnnotation(node)) {
                         TreePath declaration = getCurrentPath().getParentPath().getParentPath();
 
-                        reportedFixes.getOrReportNew(declaration, state, ReportedFixCache.NOT_AN_ERRORPRONE);
+                        reportedFixes.getOrReportNew(
+                                declaration, state, suppression -> !AllErrorprones.allBugcheckerNames(state)
+                                        .contains(suppression));
```
aldexis, line 45: You were previously removing the for-rollout prefix and aren't doing so anymore. Is that intentional?

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -83,13 +85,23 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+            Set<String> allNames =
+                    AllErrorprones.allNames(visitorState, description.checkName).get();
+            // Use the existing suppression, rather than changing it to the canonical suppression
+            String existingSuppression = AnnotationUtils.annotationStringValues(
+                            SuppressWarningsUtils.getSuppressWarnings(declaration)
+                                    .get())
+                    .filter(suppression -> allNames.contains(SuppressWarningsUtils.stripForRollout(suppression)))
+                    .collect(MoreCollectors.first())
+                    .get();
+            FIXES.getExisting(declaration).addSuppression(existingSuppression);
```
aldexis, line 104: Shouldn't you keep all the ones that might already exist, rather than just the first one? (especially the non-rollout ones)

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -83,13 +85,23 @@ public static Description interceptDescription(VisitorState visitorState, Descri
...
+            TreePath declaration = firstSuppressibleWhichSuppressesDescription.get();
+            Set<String> allNames =
+                    AllErrorprones.allNames(visitorState, description.checkName).get();
+            // Use the existing suppression, rather than changing it to the canonical suppression
+            String existingSuppression = AnnotationUtils.annotationStringValues(
+                            SuppressWarningsUtils.getSuppressWarnings(declaration)
+                                    .get())
+                    .filter(suppression -> allNames.contains(SuppressWarningsUtils.stripForRollout(suppression)))
+                    .collect(MoreCollectors.first())
+                    .get();
```
aldexis, line 103: Could this leverage `suppressionsOn` in some way?

aldexis (commented):
I tried to have a look, but I'm not familiar enough with the remove unused logic to confidently review the actual logic changes. Best to have @CRogers have an eye on this as well

suppressible-error-prone/build.gradle
CRogers: I really hope not - baseline should force Gradle to fork the compiler, with these jvm args. So shouldn't need to be added to gradle daemon via `gradle.properties`.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -16,33 +16,63 @@
...
+    private static Supplier<Map<String, Set<String>>> canonicalToAllNames = VisitorState.memoize(state -> {
+        // Use the same classloader that Error Prone was loaded from to avoid classloader skew
+        // when using Error Prone plugins together with the Error Prone javac plugin.
+        JavacProcessingEnvironment processingEnvironment = JavacProcessingEnvironment.instance(state.context);
+        ClassLoader loader = processingEnvironment.getProcessorClassLoader();
+        List<BugChecker> pluginChecks = ServiceLoader.load(BugChecker.class, loader).stream()
+                .map(Provider::get)
+                .collect(Collectors.toList());
+        EntryStream<String, Set<String>> pluginCheckNames =
+                StreamEx.of(pluginChecks).mapToEntry(BugChecker::canonicalName, BugChecker::allNames);
```
CRogers, line 46: `BugChecker#allNames` will return `for-rollout:Name` if we've patched the errorprone classes in the artifact transform.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
```diff
@@ -16,33 +16,63 @@
...
-    private static Supplier<Set<String>> cachedAllBugcheckerNames = Suppliers.memoize(() -> {
-        Stream<BugChecker> pluginChecks =
-                ServiceLoader.load(BugChecker.class).stream().map(ServiceLoader.Provider::get);
-        Stream<String> pluginCheckNames =
-                StreamEx.of(pluginChecks).flatMap(bugchecker -> bugchecker.allNames().stream());
+    private static Supplier<Map<String, Set<String>>> canonicalToAllNames = VisitorState.memoize(state -> {
+        // Use the same classloader that Error Prone was loaded from to avoid classloader skew
+        // when using Error Prone plugins together with the Error Prone javac plugin.
+        JavacProcessingEnvironment processingEnvironment = JavacProcessingEnvironment.instance(state.context);
+        ClassLoader loader = processingEnvironment.getProcessorClassLoader();
```
CRogers, line 41: I wonder if there's any real difference between this and `state.getClass().getClassLoader()`? The latter does not require interacting with so many compiler APIs or module imports.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/AllErrorprones.java
CRogers: Which may be what we want...

suppressible-error-prone/build.gradle
aldexis: Oh nevermind, I had meant to remove this comment. I just needed to have intellij somehow add it to _something_ because it didn't find the classes

PR #300, opened 2026-01-13
Support error-prone 2.46.0 by turning the `Replacement` record non-final

kelvinou01:
## Before this PR

Concurrent with https://github.com/palantir/baseline-error-prone/pull/60

Error-prone 2.43.0 introduced a major break — java 21 is now required. 

Error-prone 2.46.0 introduced a major break — [`Replacement`](https://github.com/google/error-prone/blob/v2.46.0/check_api/src/main/java/com/google/errorprone/fixes/Replacement.java) is now a `record`. This means our [`LazyReplacement` class](https://github.com/palantir/suppressible-error-prone/blob/2.26.0/suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java) could no longer extend it. 

## After this PR

- Bytecode-manipulate the `Replacement` record into a non-final class
- Introduce an `artifact-transforms` included build containing bytecode manipulation artifact transforms. This is used 
  - by gradle-suppressible-error-prone's annotationProcessor classpath (when running SEP), and
  - suppressible-error-prone's runtimeOnly and compileOnly classpath (when compiling SEP)
- This repo now targets Java 21

==COMMIT_MSG==
Support error-prone 2.46.0 by turning the `Replacement` record non-final
==COMMIT_MSG==


artifact-transforms/src/main/java/com/palantir/suppressibleerrorprone/transform/ModifiedFile.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.suppressibleerrorprone.transform;
+
+public enum ModifiedFile {
+    BUG_CHECKER_INFO("BugCheckerInfo"),
+    VISITOR_STATE("VisitorState"),
+    SUPPRESSIBLE_TREE_PATH_SCANNER("SuppressibleTreePathScanner"),
+    REPLACEMENT("Replacement");
```
kelvinou01, line 23: Moved here from gradle-suppressible-error-prone, and added the `REPLACEMENT` variant

gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/modes/modes/RemoveRolloutMode.java
```diff
@@ -30,7 +31,7 @@ public final class RemoveRolloutMode implements Mode {
     public ModifyCheckApiOption modifyCheckApi() {
         // If we're going to remove suppressions, and possibly apply patches, we don't want to apply the custom
         // logic for for-rollout suppressions.
-        return ModifyCheckApiOption.doNotModify();
+        return ModifyCheckApiOption.mustModify(ModifiedFile.REPLACEMENT);
```
kelvinou01, line 34: maybe introduce a `mustOnlyModify` primitive?

artifact-transforms/src/main/java/com/palantir/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
```diff
@@ -148,21 +155,18 @@ public abstract static class Params implements TransformParameters {
...
-                            .envVarOrFromTestingProperty("CACHE_BUST_ERRORPRONE_TRANSFORM")
-                            .map(Boolean::parseBoolean)
-                            .orElse(false)
-                            .map(should -> should ? UUID.randomUUID().toString() : ""));
+            String cacheBust = System.getenv("CACHE_BUST_ERRORPRONE_TRANSFORM");
+            if (cacheBust != null && Boolean.parseBoolean(cacheBust)) {
+                getCacheBust().set(UUID.randomUUID().toString());
+            } else {
+                getCacheBust().set("");
+            }
```
kelvinou01: revert

suppressible-error-prone/build.gradle
```diff
@@ -50,3 +61,29 @@ tasks.withType(JavaCompile) {
...
+        .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, ArtifactTypeDefinition.JAR_TYPE)
+        .attribute(ARTIFACT_TRANSFORM_APPLIED, true)
+    transform.parameters { params ->
+        params.filesToModify.set([ModifiedFile.REPLACEMENT] as Set)
+    }
+}
+
+configurations.compileClasspath.attributes.attribute(ARTIFACT_TRANSFORM_APPLIED, true)
+configurations.testCompileClasspath.attributes.attribute(ARTIFACT_TRANSFORM_APPLIED, true)
+configurations.testRuntimeClasspath.attributes.attribute(ARTIFACT_TRANSFORM_APPLIED, true)
```
CRogers: Rather than writing this out in groovy gradle code, please can we make a plugin inside the `artifact-transforms` project then just apply this? We might even be able to share/deduplicate some of the code that lives in `gradle-suppressible-error-prone` that sets up the artifact transform

suppressible-error-prone/build.gradle
```diff
@@ -1,7 +1,18 @@
...
+}
+
+plugins {
+    id 'java-library'
+    id 'com.palantir.external-publish-jar'
+}
 
 dependencies {
+    // Add dependency on artifact-transforms to get access to ModifyErrorProneCheckApi
+    implementation 'com.palantir.suppressible-error-prone:artifact-transforms'
```
CRogers: Is this needed? This jar has a hard Gradle dep, which won't be available in the java compiler, so I don't think it could ever work?

artifact-transforms/build.gradle
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id 'java-library'
+    id 'maven-publish'
+}
```
CRogers, line 19: We're going to need the `external-publish-jar` plugin, or this won't get published properly. Either that or we include this inside `gradle-suppressible-error-prone` directly.

CRogers: Currently, I think the if this were published, `gradle-suppressible-error-prone` would have a dependency on `artifact-transforms` that would never work as this is never published. We should try with an RC before merging to verify this publishing issue properly. Another option is to just include the classes in the `gradle-suppressible-error-prone` jar using the fat jar method.

artifact-transforms/src/main/java/com/palantir/suppressibleerrorprone/transform/MakeRecordNonFinalVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+            int version, int access, String name, String signature, String superName, String[] interfaces) {
+        // Remove ACC_FINAL flag from the class access modifiers
+        int modifiedAccess = access & ~Opcodes.ACC_FINAL;
+        super.visit(version, modifiedAccess, name, signature, superName, interfaces);
+    }
+
+    @Override
+    public MethodVisitor visitMethod(
+            int access, String name, String descriptor, String signature, String[] exceptions) {
+        // Remove ACC_FINAL from equals() and hashCode() methods so they can be overridden
```
CRogers, line 44: I'm actually surprised that the `replaceWith()` and `range()` methods aren't final too...

artifact-transforms/src/main/java/com/palantir/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
CRogers: Why did you need to change this? Or are you saying you should revert this?

artifact-transforms/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+        }
+    }
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation 'org.ow2.asm:asm:9.9.1'
```
CRogers: I guess this is the downside of the includedBuild approach? Can't use GCV from the outer build?

artifact-transforms/build.gradle
CRogers: Fat jarring would also include the asm dep which is less than ideal... If we could get this to publish nicely it would be better.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
```diff
@@ -107,17 +110,31 @@ private String calculateReplacementText() {
...
                     // No @SuppressWarnings, we want to prefix a new one before the start of the tree
                     int startPosition = ((DiagnosticPosition) tree).getStartPosition();
                     return Range.closedOpen(startPosition, startPosition);
                 });
     }
+
+    // Since the parent is a record, whose equals and hashCode methods solely rely on the record's fields,
+    // We have to re-override these methods to use the default implementation from Object
+    // Or else, only one Replacement will be added to any given file:
+    // https://github.com/google/error-prone/blob/a5a718974dd7d325025ea14c1492f113490d5cf8/check_api/src/main/java/com/google/errorprone/fixes/Replacements.java#L146
```
CRogers, line 133: I would add a note at the top of the file about how we're managing to extend `Replacement` despite it being final.

artifact-transforms/src/main/java/com/palantir/suppressibleerrorprone/transform/ModifyErrorProneCheckApi.java
kelvinou01: yep it was just a self-reminder to revert this

suppressible-error-prone/build.gradle
kelvinou01: yep!

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/LazySuppressionReplacement.java
kelvinou01: yep

artifact-transforms/build.gradle
kelvinou01: yeah, the included build will have it's independent resolved classpath from GCV

kelvinou01: there's no automated dependency bumping for inner versions.props currently

artifact-transforms/build.gradle
kelvinou01: yep, let's publish this as a jar and gradle plugin

CRogers:
:+1: for RC

suppressible-error-prone/build.gradle
```diff
@@ -1,5 +1,16 @@
 apply plugin: 'java-library'
 apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'com.palantir.suppressible-error-prone.setup-pre-compilation-bytecode-manipulation'
+
+buildscript {
```
CRogers: Subproject buildscripts are really prone to error (there are Gradle bugs where if different subprojects have different buildscripts, they have different classloaders for some types which can break `BuildService`s), I would just add this to the root buildscript.

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
```diff
@@ -53,6 +52,10 @@ public static Description interceptDescription(VisitorState visitorState, Descri
             SuppressionRemover.removeAllSuppressionsOnErrorprones(FIXES, compilationUnit, visitorState);
         }
 
+        if (description == Description.NO_MATCH) {
+            return description;
+        }
```
CRogers, line 57: Why has this changed?

gradle-suppressible-error-prone/build.gradle
```diff
@@ -12,6 +26,8 @@ dependencies {
     implementation 'one.util:streamex'
     implementation 'com.google.guava:guava'
 
+    shade 'com.palantir.suppressible-error-prone.internal:artifact-transforms'
```
CRogers: If we just leave it as a regular dep, can we publish the `artifact-transforms` jar in the build somehow?

CRogers: (to avoid the far jar situation)

suppressible-error-prone/src/main/java/com/palantir/suppressibleerrorprone/VisitorStateModifications.java
kelvinou01: This should've been here all along. In the previous version of the code, calling removeUnused on a file may result in no checks removed, if `SuppressionRemover.removeAllSuppressionsOnErrorprones(FIXES, compilationUnit, visitorState);` isn't called at all. This was causing one test failure — `errorProneRemoveUnused_removes_entire_suppress_warnings_annotation_when_all_suppressions_are_unused`. 

I don't know why it started failing in this PR though. I'll look into it more when I fix removeUnused. This failure won't affect things outside of removeUnused.

artifact-transforms/build.gradle
```diff
@@ -0,0 +1,40 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+    }
+    dependencies {
+        classpath 'com.gradle.publish:plugin-publish-plugin:2.0.0'
+        classpath 'com.palantir.gradle.externalpublish:gradle-external-publish-plugin:1.27.0'
+        classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:3.8.0'
+    }
```
CRogers, line 13: This should really have baseline, gradle-guide etc as well.

kelvinou01: yep, applied

gradle-suppressible-error-prone/build.gradle
kelvinou01: i've removed fat jarring in favor of publishing the included build

CRogers:
:+1:

PR #357, opened 2026-02-23
Publish artifact-transform

kelvinou01:
## Before this PR

https://github.com/palantir/suppressible-error-prone/pull/300 is broken because artifact-transform wasn't being published

## After this PR

Publish artifact-transform

==COMMIT_MSG==
Publish artifact-transform

==COMMIT_MSG==


build.gradle
```diff
@@ -12,11 +12,11 @@ buildscript {
         classpath 'com.palantir.gradle.failure-reports:gradle-failure-reports:1.18.0'
         classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:3.13.0'
         classpath 'com.gradle.publish:plugin-publish-plugin:2.0.0'
-        classpath 'com.palantir.baseline-error-prone:gradle-baseline-error-prone:0.10.0'
+        classpath 'com.palantir.baseline-error-prone:gradle-baseline-error-prone:0.9.0'
         classpath 'com.palantir.baseline:gradle-baseline-java:7.2.0'
         classpath 'com.palantir.javaformat:gradle-palantir-java-format:2.88.0'
         classpath 'com.palantir.suppressible-error-prone:gradle-suppressible-error-prone:2.29.0'
-        classpath 'com.palantir.suppressible-error-prone.internal:artifact-transforms'
+        classpath 'com.palantir.suppressible-error-prone.internal:artifact-transforms:2.30.0'
```
kelvinou01: i wonder if some special bootstrapping is needed to add [this `implementation` dep](https://github.com/palantir/suppressible-error-prone/blob/develop/gradle-suppressible-error-prone/build.gradle#L14) to versions.props?

artifact-transforms/build.gradle
```diff
@@ -35,12 +37,15 @@ gradlePlugin {
...
 
-group = 'com.palantir.suppressible-error-prone.internal'
 
 repositories {
-    mavenCentral()
+    mavenCentral() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
+    gradlePluginPortal()
 }
 
+group = 'com.palantir.suppressible-error-prone.internal'
```
CRogers: I would just keep it as the same group as everything else (ie no `internal`).

suppressible-error-prone/build.gradle
```diff
@@ -1,6 +1,9 @@
+plugins {
+    id 'com.palantir.external-publish-jar'
+    id 'com.palantir.suppressible-error-prone.setup-pre-compilation-bytecode-manipulation'
```
kelvinou01, line 3: move here so we can use plugin from included build

also move external-publish-jar, as the bytecode manip plugin requires artifact type `jar`

build.gradle
```diff
@@ -12,11 +12,10 @@ buildscript {
         classpath 'com.palantir.gradle.failure-reports:gradle-failure-reports:1.18.0'
         classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:3.13.0'
         classpath 'com.gradle.publish:plugin-publish-plugin:2.0.0'
-        classpath 'com.palantir.baseline-error-prone:gradle-baseline-error-prone:0.10.0'
+        classpath 'com.palantir.baseline-error-prone:gradle-baseline-error-prone:0.9.0'
         classpath 'com.palantir.baseline:gradle-baseline-java:7.2.0'
         classpath 'com.palantir.javaformat:gradle-palantir-java-format:2.88.0'
-        classpath 'com.palantir.suppressible-error-prone:gradle-suppressible-error-prone:2.29.0'
-        classpath 'com.palantir.suppressible-error-prone.internal:artifact-transforms'
+        classpath 'com.palantir.suppressible-error-prone:gradle-suppressible-error-prone:2.28.0'
```
kelvinou01, line 18: downgrading because 2.29.0 is looking for artifact-transforms with version `undefined`

CRogers (approved):
🙏🏻

